PR #4, opened 2024-01-09
initial graph network block

francescofarina:
Introducing the generic graph network block.

Ideally, we'll make the node, edge and global models standalone once scatter ops are merged into `mlx`.

### DIFF
--- a/examples/graph_network_block.py
+++ b/examples/graph_network_block.py
@@ -1,0 +1,161 @@
+import mlx.core as mx
+from mlx.nn.base import Module
+from mlx.nn.layers.linear import Linear
+from mlx.nn.layers.gnn import GraphNetworkBlock
+
+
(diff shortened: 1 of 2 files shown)


TristanBilot:
Now that you created an example file, you can delete the code-block in comments within the class. It will be much cleaner!

francescofarina:
I'd probably leave the code-block as that'd be part of the docs while the example will not. 

Ideally the example will be a working one (not just an instantiation+forward pass) once scatter_ops are released and we can implement efficient Node, Edge and Global models. 

What do you think?

TristanBilot:
ok I agree, anyway we'll modify everything when writing the docs

TristanBilot:
if you think it's better to remove the example file for the moment then, you can!

PR #10, opened 2024-01-09
License, packaging and pre commit hooks

francescofarina:
This PR is branched off https://github.com/TristanBilot/mlx-graphs/pull/9 and adds pre-commit hooks.

The idea is to have them run through CI/CD but until that's in place let's install them locally to ensure code quality.


TristanBilot:
thanks for the PR @francescofarina.
There's one thing I worry about using poetry instead of pip: during our development, we'll be constantly using specific branches of the original mlx repo and also local version of mlx modified with `pip install -e .`, for instance to try some new code within mlx called in mlx-graphs. As mlx uses pip and not poetry, I'm not sure it's the best thing to use poetry (?)
For example, to make the message passing work now in mlx-graphs, we need to install locally the `scatter-vjp` branch of mlx to have access to scattering operations, which are not included yet in the official mlx package. I tried to install a local version of mlx using poetry with `poetry add '/path/to/local_mlx'` and `poetry install` but it still can't find the installation of mlx. Any ideas?

TristanBilot:
Did you experiment this situation of using local packages installed with pip within poetry @gboduljak?

francescofarina:
There should be an editable install mode for poetry as well if I remember correctly. But let's double check

gboduljak:
> thanks for the PR @francescofarina. There's one thing I worry about using poetry instead of pip: during our development, we'll be constantly using specific branches of the original mlx repo and also local version of mlx modified with `pip install -e .`, for instance to try some new code within mlx called in mlx-graphs. As mlx uses pip and not poetry, I'm not sure it's the best thing to use poetry (?) For example, to make the message passing work now in mlx-graphs, we need to install locally the `scatter-vjp` branch of mlx to have access to scattering operations, which are not included yet in the official mlx package. I tried to install a local version of mlx using poetry with `poetry add '/path/to/local_mlx'` and `poetry install` but it still can't find the installation of mlx. Any ideas?

You can modify your Poetry config to install from git repo at a particular branch.
https://python-poetry.org/docs/dependency-specification/#git-dependencies

francescofarina:
perhaps the main `pyproject.toml` should contain the config for building and publishing the package, so I'm not sure having the mlx URL there is optimal.

TristanBilot:
I spent some time trying to add a separate dev group, however it doesn't work as the mlx version from the repo with URL is something like `0.0.7.dev2024110+c92a134` and the version of the original mlx has to be in range `>=0.0.7,<0.0.8`. Idk why but poetry can't figure out that the version from github actually fits between 0.07 and 0.08 so it doesn't install the env. In the doc they say that all versions have to be compatible between groups: "All dependencies must be compatible with each other across groups since they will be resolved regardless of whether they are required for installation or not (see [Installing group dependencies](https://python-poetry.org/docs/master/managing-dependencies/#installing-group-dependencies))."

TristanBilot:
I pushed my code, idk if you have an idea on how we can integrate 2 distinct dependencies for `dev` and `prod`?
Should be nice to install in dev mode with `poetry install --only dev` but can't work now due to the error I just described.

TristanBilot:
Hmm I see, we'll have to find a better solution for that. 
At the moment, we could find a trick to make it work locally for our developments, and propose a better alternative when releasing a v.0 for a production-ready env install?
If we don't merge the poetry code within main soon, it will block us for the developments

gboduljak:
> Hmm I see, we'll have to find a better solution for that. At the moment, we could find a trick to make it work locally for our developments, and propose a better alternative when releasing a v.0 for a production-ready env install? If we don't merge the poetry code within main soon, it will block us for the developments

Perhaps we should go with poetry option until we find a better alternative.

TristanBilot:
you mean we should go with poetry or without poetry?
I think now that we installed it we can keep it, but now it's not useful to keep the dependency to the original mlx as we don't use this one for dev. In few weeks mlx will be updated and we will put back the original dep.

gboduljak:
> you mean we should go with poetry or without poetry? I think now that we installed it we can keep it, but now it's not useful to keep the dependency to the original mlx as we don't use this one for dev. In few weeks mlx will be updated and we will put back the original dep.

Since we often need to switch development versions of MLX, it might be more useful to drop poetry and just use ``pip + pyproject.toml``. I feel this is not the best option forever, but could be the most convenient until we know the best option. Perhaps we should ask ``mlx`` authors for the advice here.

francescofarina:
So @TristanBilot @gboduljak I removed the poetry frontend. No we're just using it's backend to configure and build the project but venvs and requirements needs to managed manually (see updated README). This should solve all the issues above.

TristanBilot:
Much cleaner, thanks @francescofarina ! I just fixed the -e in the dev file as I think it should be @ for the git URL (-e not working for me with the URL)

TristanBilot:
However, I can't make the local install of `mlx-graphs` work for development (`pip install -e .` does not work)

README.md
```diff
@@ -17,8 +17,9 @@ Create a new virtual environment and install the requirements
 ```
 pip install -r requirements.txt
 ```
-For dev, we suggest using a different environment and use the `requirements-dev.txt`
+For dev, install `mlx-graphs` locally and install `requirements-dev.txt`
 ```
+poetry install
```
francescofarina, line 22: @TristanBilot we're not using poetry anymore, so this one shouldn't be needed

francescofarina:
@TristanBilot I completely removed poetry now and moved to the good old setuptools. Now the editable install via `pip install -e .` should work smoothly for everyone.

TristanBilot:
Indeed sorry, I was confused by the poetry mentions in `pyproject.toml`
Now it works perfectly on my side :)

TristanBilot:
If ok for everyone, we can merge

PR #11, opened 2024-01-11
GAT, scatter_softmax, and tests

TristanBilot:
After implementing [scatter_max vjp](https://github.com/ml-explore/mlx/pull/431) (merged soon in `mlx.core`), we can start implementing GAT 🚀

This is just a draft, tests fail yet, I need more time to make the shapes work with broadcasting and stuff.
Nevermind, you can take a look in the meantime!

In the future, it may be interesting to implement the overall scatter_softmax in plain C++ using scatter sum and max operations jointly to improve the perfs, but the aim of this PR is just to have a working GATConv with proper shapes.

PS: sorry for the large PR 🙏


gboduljak:
I can implement ``scatter_softmax`` in C++ on ``mlx`` repo soon.

TristanBilot:
Great! I can work alongside you if needed

mlx_graphs/utils/scatter.py
```diff
@@ -0,0 +1,46 @@
...
+    if aggr == "max":
+        return scatter_max(empty_tensor, indices, values)
+
+
+def scatter_add(src: mx.array, indices: mx.array, values: mx.array, axis: int = 0):
+    return mx.scatter_add(src, indices, values, axis)
+
+
+def scatter_max(src: mx.array, indices: mx.array, values: mx.array, axis: int = 0):
+    return mx.scatter_max(src, indices, values, axis)
```
francescofarina, line 30: I'm not sure where we stand on this in `mlx`. Are we back to having the `scatter_{ops}` implemented? I understood we were going with the `.at[]` syntax.

mlx_graphs/utils/utils.py
```diff
@@ -0,0 +1,14 @@
...
+
+import mlx.core as mx
+
+
+def gather_src_dst(x: mx.array, edge_index: mx.array) -> Tuple[mx.array, mx.array]:
+    src_idx, dst_idx = edge_index
+    x_i = x[src_idx]
+    x_j = x[dst_idx]
+
+    return x_i, x_j
```
francescofarina, line 11: I suggest using more readable function and variable names. Right now, this could be interpreted as just gathering indices, while that's not what it's doing.

```suggestion
def map_node_features_to_index(x: mx.array, edge_index: mx.array) -> Tuple[mx.array, mx.array]:
    """
    Extracts source and destination node features based on the given edge indices.

    Parameters:
    - x (mx.array): The input array of node features.
    - edge_index (mx.array): An array of size [2, number_of_edges], where each columns contains the source
                and destination nodes of an edge.

    Returns:
    Tuple[mx.array, mx.array]: A tuple containing source and destination features.
    """
    source_indices, destination_indices = edge_index
    source_features = features[source_indices]
    destination_features = features[destination_indices]

    return source_features, destination_features
```

mlx_graphs/nn/conv/gat_conv.py
```diff
@@ -0,0 +1,98 @@
...
+from mlx_graphs.nn.message_passing import MessagePassing
+from mlx_graphs.utils import glorot_init, scatter_softmax, gather_src_dst
+
+
+class GATConv(MessagePassing):
+    r"""Graph Attention Network convolution layer.
+
+    Args:
+        x_dim (int): size of input node features
+        h_dim (int): size of hidden node embeddings
```
francescofarina, line 16: these should probably be `input` and `output` dimensions instead of `x` and `h`

mlx_graphs/nn/conv/gat_conv.py
```diff
@@ -0,0 +1,98 @@
...
+
+        if "bias" in self:
+            h = h + self.bias
+
+        return h
+
+    def message(
+        self, x_i: mx.array, x_j: mx.array, alpha: mx.array=None, **kwargs: Any
+    ) -> mx.array:
+        return alpha.reshape(*alpha.shape, 1) * x_i
```
francescofarina, line 84: `i`, `j` -> `src`, `dst`

mlx_graphs/nn/conv/gat_conv.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _edge_update(self, alpha_j: mx.array, alpha_i: mx.array, index: mx.array):
```
francescofarina, line 86: index -> edge_index?

mlx_graphs/nn/message_passing.py
```diff
@@ -25,27 +28,37 @@ def __init__(self, aggr="add"):
         super().__init__()
 
         self.aggr = aggr
+        self.node_dim = None
 
     def __call__(self, x: mx.array, edge_index: mx.array, **kwargs: Any):
         raise NotImplementedError
 
-    def propagate(self, x: mx.array, edge_index: mx.array, **kwargs: Any) -> mx.array:
+    def propagate(self, x: Union[mx.array, ArrayTuple], edge_index: mx.array, **kwargs: Any) -> mx.array:
```
francescofarina, line 36: I'm not sure about the utility of the `ArrayTuple`, probably better to stick to `mx.array`

mlx_graphs/utils/inits.py
francescofarina, line 1: I'm not sure this should be part of `mlx-graphs`

francescofarina (changes requested):

mlx_graphs/utils/scatter.py
TristanBilot: You're right, I will not let these, just using them locally in my mlx code that's why. Will remove them once scatter_max vjp is merged in mlx

mlx_graphs/utils/utils.py
TristanBilot: True! I'll modify that

mlx_graphs/nn/conv/gat_conv.py
TristanBilot: Yes, as discussed earlier we will not use x and h anymore, so I'll change that too

mlx_graphs/nn/conv/gat_conv.py
TristanBilot: No, here we perform the update wrt to either src or dst index, meaning that we don't use both like in edge index. I will explain that in the docs.

mlx_graphs/utils/inits.py
TristanBilot: I created this only for GAT implementation. The thing is that glorot init is not implemented directly in MLX, the default init of Linear is not exactly the same. Experiments will tell us whether the default Linear init is enough or we must add our own glorot

mlx_graphs/nn/message_passing.py
TristanBilot: Me too, I'll remove it now and we'll see in the near future if some long typings are frequently used in the project

mlx_graphs/nn/conv/gat_conv.py
francescofarina: ok, great!

mlx_graphs/nn/conv/gat_conv.py
```diff
@@ -0,0 +1,135 @@
...
+    r"""Graph Attention Network convolution layer.
+
+    Args:
+        x_dim (int): size of input node features
+        h_dim (int): size of hidden node embeddings
+        heads (int): number of attention heads
+        concat (bool): whether to use concat of heads or mean reduction
+        bias (bool): whether to use bias in the node projection
+        negative_slope (float): slope for the leaky relu
+        dropout (float): probability p for dropout
```
francescofarina, line 21:
```suggestion
        dropout (float): probability p for dropout
        edge_features_dim (int, optional): size of the edge features
```

mlx_graphs/utils/utils.py
```diff
@@ -0,0 +1,72 @@
...
+            src = mx.expand_dims(src, 0)
+    for _ in range(src.ndim, other.ndim):
+        src = mx.expand_dims(src, -1)
+    src = expand(src, other.shape)
+    return src
+
+
+def expand(array: mx.array, new_shape: tuple) -> mx.array:
+    """
+    Expand the dimensions of a NumPy array to a new shape.
```
francescofarina, line 56:
```suggestion
    Expand the dimensions of an array to a new shape.
```

examples/gat.py
francescofarina, line 1: I'm wondering whether it'd make sense to start benchmarking against other libraries to see performance. I guess we still need to do a bit more work before that but still

examples/gcn.py
```diff
@@ -19,19 +22,21 @@ class GCN(nn.Module):
 
     def __init__(
         self,
-        x_dim: int,
-        h_dim: int,
-        out_dim: int,
-        nb_layers: int = 2,
+        node_features_dim: int,
+        hid_features_dim: int,
```
francescofarina, line 26: hid -> hidden?

mlx_graphs/nn/conv/gat_conv.py
```diff
@@ -0,0 +1,158 @@
...
+        edge_features_dim: Optional[int] = None,
+    ):
+        super(GATConv, self).__init__(aggr="add")
+
+        self.out_features_dim = out_features_dim
+        self.heads = heads
+        self.concat = concat
+        self.negative_slope = negative_slope
+
+        # NOTE: Check to add glorot_init within the Linear layer
```
francescofarina, line 51: rm?

mlx_graphs/nn/message_passing.py
```diff
@@ -11,7 +13,7 @@ class MessagePassing(nn.Module):
     passing paradigm. This implementation is inspired from PyTorch Geometric [2].
 
     Args:
-        aggr (str): the aggregation strategy used to aggregate messages
+        aggr (str): Aggregation strategy used to aggregate messages
```
francescofarina, line 16: Please add the available options.

Also, should `aggr` be a `Literal` with the different available options rather than a string? Even an `Enum` but that may add some extra burden to the UX.

mlx_graphs/nn/message_passing.py
```diff
@@ -25,72 +27,89 @@ def __init__(self, aggr="add"):
...
         raise NotImplementedError
 
-    def propagate(self, x: mx.array, edge_index: mx.array, **kwargs: Any) -> mx.array:
+    def propagate(
+        self,
+        node_features: Union[mx.array, Tuple[mx.array, mx.array]],
+        edge_index: mx.array,
+        message_kwargs: Optional[Dict] = {},
+        aggregate_kwargs: Optional[Dict] = {},
+        update_kwargs: Optional[Dict] = {},
```
francescofarina, line 41: I know this is how also PyG implements it, but is there a better way to expose them?

mlx_graphs/utils/scatter.py
```diff
@@ -0,0 +1,129 @@
...
+    Parameters:
+        src (mx.array): Source array where the values will be scattered (often an empty array)
+        index (mx.array): Array containing indices that determine the scatter of the 'values'.
+        values (mx.array): Input array containing values to be scattered.
+
+    Returns:
+        mx.array: The resulting array after applying scatter and max operations on the values
+            at duplicate indices
+    """
+    return src.at[index].maximum(values)
```
francescofarina, line 93: do we really need these if the design choice of `mlx` is to use the `.at`?

mlx_graphs/utils/scatter.py
```diff
@@ -0,0 +1,129 @@
...
+    axis: Optional[int] = 0,
+) -> mx.array:
+    """
+    Default function for performing all scattering operations.
+    Scatters `values` at `index` in an empty array of `out_size` elements.
+
+    Args:
+        values (mx.array): array with all the values to scatter in the output tensor
+        index (mx.array): array with index to which scatter the values
+        out_size (optional, int): number of elements in the output array (size of the first dimension).
```
francescofarina, line 22: `optional, int` -> `int, optional` 

there are a couple of other ones below

francescofarina (changes requested):

mlx_graphs/nn/message_passing.py
```diff
@@ -25,72 +27,89 @@ def __init__(self, aggr="add"):
...
+    ) -> mx.array:
         r"""Computes messages from neighbors, aggregates them and updates
         the final node embeddings.
 
         Args:
-            x (mx.array): input node features/embeddings
-            edge_index (mx.array): graph representation of shape (2, |E|) in COO format
-            **kwargs (Any): arguments to pass to message, aggregate and update
+            node_features (Union[mx.array, Tuple[mx.array, mx.array]]): Input node features/embeddings
```
francescofarina, line 47: Shall we make this always a `mx.array`?

mlx_graphs/nn/message_passing.py
```diff
@@ -25,72 +27,89 @@ def __init__(self, aggr="add"):
...
+        if isinstance(node_features, tuple):
+            assert (
+                len(node_features) == 2
+                and isinstance(node_features[0], mx.array)
+                and isinstance(node_features[1], mx.array)
+            ), "Invalid shape for `node_features`, should be a tuple of 2 mx.array."
+        else:
+            assert isinstance(
+                node_features, mx.array
+            ), f"Invalid shape for `node_features`, should be an `mx.array`, found {type(node_features)}."
```
francescofarina, line 65: Can you make all these explicitly raise a `ValueError`?

francescofarina (changes requested):

TristanBilot:
Quick update:

* the NaN values were due to the `log_softmax`, not GAT.
* I noticed a significant difference between using the default Linear of mlx and a Linear with the same glorot init of PyG. Indeed, the latter considers both input and output dims for the normalization, instead of mlx's Linear that only considers input dims. For GNNs, I think it's more suitable to use this normalization as gradients can grow quickly with large aggregations. On GAT, it improves test accuracy by 4%, but on GCN it seems to be worst (I need further analysis on that). Anyway, I think it's good idea to add it within mlx-graphs even if it's not the best place to put it.

francescofarina:
> Quick update:
> 
> * the NaN values were due to the `log_softmax`, not GAT.
> * I noticed a significant difference between using the default Linear of mlx and a Linear with the same glorot init of PyG. Indeed, the latter considers both input and output dims for the normalization, instead of mlx's Linear that only considers input dims. For GNNs, I think it's more suitable to use this normalization as gradients can grow quickly with large aggregations. On GAT, it improves test accuracy by 4%, but on GCN it seems to be worst (I need further analysis on that). Anyway, I think it's good idea to add it within mlx-graphs even if it's not the best place to put it.

@TristanBilot looks like we're starting to implement additional initializers in `mlx` (including glorot) https://github.com/ml-explore/mlx/pull/456

TristanBilot:
> @TristanBilot looks like we're starting to implement additional initializers in `mlx` (including glorot) [ml-explore/mlx#456](https://github.com/ml-explore/mlx/pull/456)

Cool! I suggest that we keep the glorot init locally now and we remove it once the PR on inits is merged.

mlx_graphs/nn/conv/gat_conv.py
```diff
@@ -0,0 +1,157 @@
...
+            )
+        else:
+            node_features = mx.mean(node_features, axis=1)
+
+        if "bias" in self:
+            node_features = node_features + self.bias
+
+        return node_features
+
+    def message(
```
francescofarina, line 122: Can you add som docs?

francescofarina (approved):
Cool - just one extra minor comment on adding some docs for the `message` method. I think we can merge.

PR #13, opened 2024-01-14
Initial adjacency matrix transformations

francescofarina:
This PR implements:
- `to_edge_index`: to get the edge_index starting from a dense adjacency matrix
- `to_sparse_adjacency_matrix`: to go from a dense adjacency matrix to a sparse representation (edge_index, edge_features)
- `to_adjacency_matrix`: inverse of the above

Notes: 
- using `numpy` as `mlx` doesn't directly implement `nonzero`
- I'm also temporarily removing pyright from the committed pre-commit as it needs to be configured with the local venv to work.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,10 +7,6 @@ repos:
       args: [--fix]
     # Run the formatter.
     - id: ruff-format
-- repo: https://github.com/RobertCraigie/pyright-python
-  rev: v1.1.341
-  hooks:
-    - id: pyright
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v2.3.0
   hooks:
--- a/mlx_graphs/utils/transformations.py
+++ b/mlx_graphs/utils/transformations.py
@@ -1,3 +1,4 @@
+from typing import Optional
 import mlx.core as mx
 import numpy as np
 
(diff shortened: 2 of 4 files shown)


mlx_graphs/utils/transformations.py
```diff
@@ -0,0 +1,81 @@
...
+        [mx.array(x, dtype=input_dtype) for x in np.nonzero(adjacency_matrix)]
+    )
+    return edge_index
+
+
+@check_adjacency_matrix
+def to_sparse_adjacency_matrix(adjacency_matrix: mx.array) -> tuple[mx.array, mx.array]:
+    """
+    Converts a dense adjacency matrix to a sparse one, represented as an tuple of and
+    edge index and edge features.
```
TristanBilot, line 56: represented as a tuple of edge index and edge features*

mlx_graphs/utils/transformations.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    input_dtype = adjacency_matrix.dtype
+    edge_index = mx.stack(
+        [mx.array(x, dtype=input_dtype) for x in np.nonzero(adjacency_matrix)]
+    )
+    return edge_index
+
+
+@check_adjacency_matrix
+def to_sparse_adjacency_matrix(adjacency_matrix: mx.array) -> tuple[mx.array, mx.array]:
```
TristanBilot, line 53: Isn't `sparse` ambiguous here? A sparse adj matrix usually refers to a specific sparse data type like a `SparseTensor` in PyG. We'll certainly implement these sparse types in the future so maybe using here `to_edge_index_with_features` would be less ambiguous?

tests/test_transformations.py
```diff
@@ -0,0 +1,49 @@
...
+        with pytest.raises(expected_exception):
+            foo(adjacency_matrix=x)
+    else:
+        assert foo(adjacency_matrix=x) is True
+
+
+def test_to_edge_index():
+    matrix = mx.array([[0, 1, 2], [3, 0, 0], [5, 1, 2]])
+    edge_index = to_edge_index(matrix)
+    assert edge_index.dtype == matrix.dtype
```
TristanBilot, line 35: To be consistent in all tests, we have to determine whether we use a descriptive message in the `assert` tests. What do you think about that?

TristanBilot: I personally used assert messages in the GAT PR as they provide some context in those tests, but I'm not sure it will be necessary in all tests.

tests/test_transformations.py
TristanBilot, line 1: For consistency, can you move the file in a `tests/utils/` folder?

.pre-commit-config.yaml
TristanBilot, line 1: Yes we definitely have to set this up

mlx_graphs/utils/transformations.py
francescofarina: I'm not sure that's ambiguous here as it's pretty much how a sparse matrix is defined/represented. Also looking at PyG, it seems their [`dense_to_sparse`](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/utils/sparse.html#dense_to_sparse) is pretty similar and doesn't necessarily imply the use of sparse tensors. The `adjacency_matrix` input may be a sparse tensor once they're implemented, but the edge index and features that this function outputs won't be sparse by definition (as they'll store information about nonzero elements of the input `adjacency_matrix`).

@TristanBilot what do you think?

tests/test_transformations.py
francescofarina: Yes, let's do that as much as we can.

mlx_graphs/utils/transformations.py
TristanBilot: If it's implemented similarly in PyG then go for it!

TristanBilot (approved):
🚀

PR #16, opened 2024-01-17
Data class

francescofarina:


mlx_graphs/data/data.py
```diff
@@ -0,0 +1,61 @@
...
+        **kwargs: Additional keyword arguments to store any other custom attributes.
+    """
+
+    def __init__(
+        self,
+        edge_index: Optional[mx.array] = None,
+        node_features: Optional[mx.array] = None,
+        edge_features: Optional[mx.array] = None,
+        graph_features: Optional[mx.array] = None,
+        node_labels: Optional[mx.array] = None,
```
TristanBilot, line 35: Is it really necessary to have labels for all of them?
I don't see a use case where we want to do classification of node/edges at the same time.

mlx_graphs/data/data.py
```diff
@@ -0,0 +1,61 @@
...
+        """Converts the Data object to a dictionary.
+
+        Returns:
+            dict: A dictionary representation of the Data object.
+        """
+        return self.__dict__
+
+
+class Dataset:
+    def __init__(self, data: list[Data]) -> None:
```
TristanBilot, line 60: -> None not necessary in __init__

mlx_graphs/data/data.py
```diff
@@ -0,0 +1,61 @@
...
+        self.dataset = dx.buffer_from_vector(data)  # type: ignore
```
TristanBilot, line 61: Did you find documentation about `buffer_from_vector`?
I'm not sure this allows to load `Data` lazily one by one.

TristanBilot: It looks like `Data` has to be iterable, got this error: `'Data' object is not iterable
`

TristanBilot: Just found the doc [here](https://ml-explore.github.io/mlx-data/build/html/python/_autosummary/mlx.data.buffer_from_vector.html#mlx.data.buffer_from_vector).

mlx_graphs/data/data.py
francescofarina: I think it makes sense to have them there as they're all optional. At least to have a base default structure.

mlx_graphs/data/data.py
francescofarina: Yes, that was WIP - should be ok now.

However, I'm testing it and
```
self.dataset = dx.buffer_from_vector([x.to_dict() for x in data])
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: contiguous array expected -- use numpy.ascontiguousarray()
```

francescofarina: > Did you find documentation about `buffer_from_vector`? I'm not sure this allows to load `Data` lazily one by one.

On this - from the docs it seems that Data should be loaded lazily https://ml-explore.github.io/mlx-data/build/html/python/_autosummary/mlx.data.buffer_from_vector.html

TristanBilot: Ok then we have to provide 2 options:
* load a dataset from a list of `Data`, this will be not lazy as the list of `Data` is already in memory.
* load a dataset from a list of files that store `Data` objects, this will loads each `Data` lazily from disk.

francescofarina: yes, that's an option. The only issue I see with that is that storing `Data` objects as files may not be optimal. Perhaps there's a way to automatically do that when you pass a list of `Data`.

francescofarina: BTW: it seems that `mlx.data` requires to work with `numpy` arrays instead of `mlx` arrays.

TristanBilot: > BTW: it seems that `mlx.data` requires to work with `numpy` arrays instead of `mlx` arrays.

Hmm ok strange

TristanBilot: > yes, that's an option. The only issue I see with that is that storing `Data` objects as files may not be optimal. Perhaps there's a way to automatically do that when you pass a list of `Data`.

I'm pretty sure that it's the only solution for real lazy loading. Take this example: you have a dataset with 10 graphs of size 10GB. Initializing `Dataset` with the 10 `Data` is not feasible. Instead, we have either to load each `Data` directly from a kind of pkl file or to download it from internet and then load the `Data`.

francescofarina: > > yes, that's an option. The only issue I see with that is that storing `Data` objects as files may not be optimal. Perhaps there's a way to automatically do that when you pass a list of `Data`.
> 
> I'm pretty sure that it's the only solution for real lazy loading. Take this example: you have a dataset with 10 graphs of size 10GB. Initializing `Dataset` with the 10 `Data` is not feasible. Instead, we have either to load each `Data` directly from a kind of pkl file or to download it from internet and then load the `Data`.

Yes, sure, in that case it makes sense - but perhaps we should have two different classes for the two cases?

francescofarina: > > BTW: it seems that `mlx.data` requires to work with `numpy` arrays instead of `mlx` arrays.
> 
> Hmm ok strange

This works
```
dx.buffer_from_vector([{"x": np.zeros([1, 1])} for i in range(10)])
```
This raises a `ValueError`
```
dx.buffer_from_vector([{"x": mx.zeros([1, 1])} for i in range(10)])
```

I'll ask in `mlx-data` whether that's desired behavior.

TristanBilot: > Yes, sure, in that case it makes sense - but perhaps we should have two different classes for the two cases?

Maybe

francescofarina: > > > BTW: it seems that `mlx.data` requires to work with `numpy` arrays instead of `mlx` arrays.
> > 
> > 
> > Hmm ok strange
> 
> This works
> 
> ```
> dx.buffer_from_vector([{"x": np.zeros([1, 1])} for i in range(10)])
> ```
> 
> This raises a `ValueError`
> 
> ```
> dx.buffer_from_vector([{"x": mx.zeros([1, 1])} for i in range(10)])
> ```
> 
> I'll ask in `mlx-data` whether that's desired behavior.

Answer should be [here](https://ml-explore.github.io/mlx-data/build/html/buffers_streams_samples.html#samples)

TristanBilot: We'll see if it's a good choice to use mlx-data. It seems that the advantages of using this package are mainly for image/text processing. If it just provides a wrapper class with an iterator on the data, I don't think it's worth it to add this dependency as we can create a dedicated class for that. However, it seems that it is fast to load objects from files (as shown in the example with images), so maybe this would be something to explore.

francescofarina: Ok, shall we remove it for now and perhaps reintroduce it later if needed?

TristanBilot: Yes we can do that. We can give it another try when implementing the lazy `Dataset` class for large graphs.

francescofarina: Let me leave just the `Data` class in this PR so that we can start using that one at least.

TristanBilot: @Kheims can you submit a PR with your beginning work on batching? So that we can have a look on what to do next.

TristanBilot: > Let me leave just the `Data` class in this PR so that we can start using that one at least.

Ok! So you remove the mlx-data stuff?

francescofarina: Yes, now it contains just the `Data` class, very minimal PR but we need that block anyways.

Kheims:
Wouldn't it be helpful to provide a class method that creates a Data out of a pyg's instance to ease later testing ? 

For the batching part, the Graphs class ([see Tristan proposition](https://github.com/TristanBilot/mlx-graphs/issues/6#issuecomment-1888949140)) would play the same role as the Batch in pyg, it's primary role is to provide a batch out of a list of Graph (Data) objects, **building a common big graph** (ie single edge_index and node features)  out of the passed graphs, from which we can get the according individual graph and also a slice of corresponding graphs, the main painpoint is to increment accordingly the indices as graph 1 and 2 shouldn't share the same indices so it has to be dealt with internally.

```
g1 = Graph(node_features, edge_features ...)
g2 = Graph(node_features2, edge_features2 ...)
batch  = Graphs([g1,g2]) # or we might instantiate it from a class method from_graph_list to make it more explicit 

```

Kheims:
Wouldn't it be helpful to provide a class method that creates a Data out of a pyg's instance to ease later testing ? 

For the batching part, the Graphs class ([see Tristan proposition](https://github.com/TristanBilot/mlx-graphs/issues/6#issuecomment-1888949140)) would play the same role as the Batch in pyg, it's primary role is to provide a batch out of a list of Graph (Data) objects, **building a common big graph** (ie single edge_index and node features)  out of the passed graphs, from which we can get the according individual graph and also a slice of corresponding graphs, the main painpoint is to increment accordingly the indices as graph 1 and 2 shouldn't share the same indices so it has to be dealt with internally.

```
node_features1 = mx..random.randint(1, (5,10))
g1 = Graph(node_features, edge_features ...)
g2 = Graph(node_features2, edge_features2 ...)
batch  = Graphs([g1,g2]) # or we might instantiate it from a class method from_graph_list to make it more explicit 


```

Kheims:
Wouldn't it be helpful to provide a class method that creates a Data out of a pyg's instance to ease later testing ? 

For the batching part, the Graphs class ([see Tristan proposition](https://github.com/TristanBilot/mlx-graphs/issues/6#issuecomment-1888949140)) would play the same role as the Batch in pyg, it's primary role is to provide a batch out of a list of Graph (Data) objects, **building a common big graph** (ie single edge_index and node features)  out of the passed graphs, from which we can get the according individual graph and also a slice of corresponding graphs, the main painpoint is to increment accordingly the indices as graph 1 and 2 shouldn't share the same indices so it has to be dealt with internally.

```
node_features1 = mx..random.randint(1, (5,10))
node_features2 = mx..random.randint(1, (10,10))
edge_features1 = mx..random.randint(5, (10,2))
edge_features2 = mx..random.randint(10, (20,2))
g1 = Graph(node_features, edge_features ...)
g2 = Graph(node_features2, edge_features2 ...)
batch  = Graphs([g1,g2]) # or we might instantiate it from a class method from_graph_list to make it more explicit 
batch 
# Graphs(x=[15, 5], edge_index=[20, 4], batch=[15])
```
indexing and slicing will be taken care, facilitating custom iteration and also usage in a loader if necessary.
A Dataset would then be created out of a list of Graphs.
 it is still a WIP a PR will be made soon but if you have insight or comments let me know @francescofarina @TristanBilot

francescofarina:
@Kheims thanks for this. I'm happy to rename the `Data` class to `Graph` (or maybe `GraphData` as having `Graph` and `Graphs` may create some confusion).

Some extra comments below.

> Wouldn't it be helpful to provide a class method that creates a Data out of a pyg's instance to ease later testing ?

I'm not sure relying on PyG would be optimal - perhaps what we can do is add to the `Data` class a `from_pyg_data` method? 

> For the batching part, the Graphs class ([see Tristan proposition](https://github.com/TristanBilot/mlx-graphs/issues/6#issuecomment-1888949140)) would play the same role as the Batch in pyg, it's primary role is to provide a batch out of a list of Graph (Data) objects, **building a common big graph** (ie single edge_index and node features) out of the passed graphs, from which we can get the according individual graph and also a slice of corresponding graphs, the main painpoint is to increment accordingly the indices as graph 1 and 2 shouldn't share the same indices so it has to be dealt with internally.

I'm not sure I completely get "graph 1 and 2 shouldn't share the same indices so it has to be dealt with internally". In which case would that happen? I think the indices of each graph in a batch are disjoint sets by default. Basically what the `Batch` does in PyG is to block-wise create a unique adjacency matrix from the various data points - each block in the adjacency matrix is corresponds to the adjacency matrix of a single data.

Kheims:
> @Kheims thanks for this. I'm happy to rename the `Data` class to `Graph` (or maybe `GraphData` as having `Graph` and `Graphs` may create some confusion).
 
 I agree, GraphData seems more appropriate.
 
> > Wouldn't it be helpful to provide a class method that creates a Data out of a pyg's instance to ease later testing ?
> 
> I'm not sure relying on PyG would be optimal - perhaps what we can do is add to the `Data` class a `from_pyg_data` method?

Yes, I think this would help.

> > For the batching part, the Graphs class ([see Tristan proposition](https://github.com/TristanBilot/mlx-graphs/issues/6#issuecomment-1888949140)) would play the same role as the Batch in pyg, it's primary role is to provide a batch out of a list of Graph (Data) objects, **building a common big graph** (ie single edge_index and node features) out of the passed graphs, from which we can get the according individual graph and also a slice of corresponding graphs, the main painpoint is to increment accordingly the indices as graph 1 and 2 shouldn't share the same indices so it has to be dealt with internally.
> 
> I'm not sure I completely get "graph 1 and 2 shouldn't share the same indices so it has to be dealt with internally". In which case would that happen? I think the indices of each graph in a batch are disjoint sets by default. Basically what the `Batch` does in PyG is to block-wise create a unique adjacency matrix from the various data points - each block in the adjacency matrix is corresponds to the adjacency matrix of a single data.

Indeed when having an adjacency matrix the indices are disjoint by default, but from my understanding when having a an edge representation the problem arise,  see [Issue](https://github.com/pyg-team/pytorch_geometric/issues/179) and [see](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/data/batch.html#Batch) the increment bool in the collate fn is True when dealing with edge repr

francescofarina:
> Indeed when having an adjacency matrix the indices are disjoint by default, but from my understanding when having a an edge representation the problem arise, see [Issue](https://github.com/pyg-team/pytorch_geometric/issues/179) and [see](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/data/batch.html#Batch) the increment bool in the collate fn is True when dealing with edge repr

Yes, it should be a pretty similar procedure though. Basically, what you have to do is to shift the indices in each new edge_index by the `cumsum` of the nodes in the already aggregated graphs?

For example, if you have two data points
```python
g1 = GraphData(edge_index=mx.array(([0, 1, 2], [2, 2, 1]])) # this has 3 nodes
g2 = GraphData(edge_index=mx.array(([0, 1, 2], [2, 2, 1]]))
```
when you collate them into a `Graphs` you'll have that the resulting `edge_index` will be
```
batch = Graphs([g1, g2])
batch.edge_index
# mx.array([
#            [0, 1, 2, 3, 4, 5],
#            [2, 2, 1, 5, 5, 4],
# ])                   ^
#                      |
#                      the ones from here corresponds to g2 and are computed as g2.edge_index + g1.num_nodes
```

francescofarina:
> This will be effective for datasets with multiple graphs (inductive). For transductive tasks with a single graph, we also have to find a way to sample multiple batches from the large graph.
> 
> I don't know if the batching is the same in both cases?

I guess that can be done via masking? Or perhaps it's more memory efficient to sample and then collate the samples.

PR #18, opened 2024-01-18
Add suport for inf values in GCN

Kheims:
Adding support for inf values in the computation of the norm when dealing with isolated nodes


mlx_graphs/nn/conv/gcn_conv.py
```diff
@@ -55,7 +54,17 @@ def message(
...
 
         return mx.scatter_add(out, index, one.reshape(-1, 1), 0)
+    
+    def _deg_inv_sqrt(deg: mx.array) -> mx.array:
+
+        deg += 1e-10
+        deg_inv_sqrt = deg**(-0.5)
+        deg_inv_sqrt = mx.where(deg_inv_sqrt <= 1, deg_inv_sqrt, 0)
+        
+        return deg_inv_sqrt
```
francescofarina, line 69: Can you add some docs here and use a self-explaining name for the function?

mlx_graphs/nn/conv/gcn_conv.py
```diff
@@ -55,7 +54,17 @@ def message(
...
     def _degree(self, index: mx.array, num_edges: int) -> mx.array:
+        
         out = mx.zeros((num_edges,))
         one = mx.ones((index.shape[0],), dtype=out.dtype)
 
         return mx.scatter_add(out, index, one.reshape(-1, 1), 0)
+    
+    def _deg_inv_sqrt(deg: mx.array) -> mx.array:
+
+        deg += 1e-10
```
francescofarina, line 65: I'm not sure whether hardcoding this is the best option. It should probably depend on the `dtype` of `deg`?

francescofarina (changes requested):

tests/utils/test_gcn.py
```diff
@@ -0,0 +1,11 @@
...
+import pytest
+
+from mlx_graphs.nn.conv.gcn_conv import GCNConv
+
+
+def test_inv_degree():
+    deg = mx.array([1, 2, 3, 0])
+    expected = mx.array([1, 0.707107, 0.57735, 0])
+    out = GCNConv._deg_inv_sqrt(deg)
+    assert mx.array_equal(out, expected), "Incorrect degree inversion"
```
TristanBilot, line 11: Could you add a test with an inf float?

```python
mx.array([1, 2, 3, float("inf")])
```

mlx_graphs/nn/conv/gcn_conv.py
Kheims: I agree tbh, but this is a workaround suggested by awni as boolean indexing isn't yet available [see](https://github.com/ml-explore/mlx/discussions/408)

francescofarina: Yes, what I'm saying is that `1e-10` is not necessarily the best value to use for any `dtype`. For example that's not representable by a float16.

tests/utils/test_gcn.py
Kheims: What i mean is that mlx right now has issues dealing with inf values when using mx.where so an example where there is a 0 in the list before the **-0.5 actually corresponds to an inf if we hadn't add an epsilon but i agree it's too ambiguous, will add documentation and rechange names to make it more explicit

mlx_graphs/nn/conv/gcn_conv.py
Kheims: What value would you suggest ? 1e-6 is a value that would be representable but actually it's only equal to the expected output rounding to 5 .

francescofarina: Yeah, I guess I'd make it depend on `deg.dtype`. What do you think?

mlx_graphs/nn/conv/gcn_conv.py
```diff
@@ -76,3 +75,11 @@ def message(
     def _degree(self, index: mx.array, num_edges: int) -> mx.array:
         one = mx.ones((index.shape[0],))
         return scatter(one, index, num_edges, "add")
+
+    def _deg_inv_sqrt(deg: mx.array) -> mx.array:
+        deg_type = {mx.float16: 1e-6, mx.float32: 1e-10}
+        deg += deg_type(deg.dtype)
```
TristanBilot, line 81: I don't know how this can work since your do `deg_type(deg.dtype)` instead of `deg_type[deg.dtype]`. Did you run tests?

mlx_graphs/nn/conv/gcn_conv.py
```diff
@@ -76,3 +75,11 @@ def message(
     def _degree(self, index: mx.array, num_edges: int) -> mx.array:
```
TristanBilot, line 75: the degree function has moved but I'll handle that during merge.

TristanBilot (changes requested):

mlx_graphs/utils/scatter.py
```diff
@@ -182,3 +182,24 @@ def degree(index: mx.array, num_nodes: Optional[int] = None) -> mx.array:
 
     ones = mx.ones((index.shape[0],))
     return scatter(ones, index, num_nodes, "add")
+
+
+def invert_sqrt_degree(degree: mx.array) -> mx.array:
+    """
+    Function that computes the inverted square root of the degree array.
```
TristanBilot, line 189:
```suggestion
    Computes the inverted square root of the degree array. NOTE: This is a temporary workaround to deal with infinite values according to the GCN paper as boolean indexing isn't yet available, so we have to pre-pad zero elements of the degree array (i.e. isolated nodes)
```

mlx_graphs/utils/scatter.py
```diff
@@ -182,3 +182,24 @@ def degree(index: mx.array, num_nodes: Optional[int] = None) -> mx.array:
...
+    NOTE: This is a temporary workaround to deal with infinite values according to
+        GCNPaper as boolean indexing isn't yet available so we have to pre-pad zero
+        elements of the degree array (ie isolated nodes)
+
+    Args:
+        degree: Array of length `num_nodes` with the degree of each node.
```
TristanBilot, line 195:
```suggestion
        degree: Array of length num_nodes with the inverted square root degree of each node.
```

TristanBilot:
Can you also add `invert_sqrt_degree` to `scatter.rst` so that the function is included into the docs? After that and the 2 minor comment changes, we can 🚢

Kheims:
> Can you also add `invert_sqrt_degree` to `scatter.rst` so that the function is included into the docs? After that and the 2 minor comment changes, we can 🚢

Done

PR #22, opened 2024-01-19
Initial CI

francescofarina:
Setting up initial CI workflow to test linting and formatting.

> Note: having tests (via `pytest`) in CI is currently not possible because they need to run on Apple silicon.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     strategy:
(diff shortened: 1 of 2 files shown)


TristanBilot:
Dope! Should we also integrate tests in the CI?

francescofarina:
> Dope! Should we also integrate tests in the CI?

I added them initially but they need to run on Apple Silicon and that's only available for large runners on GitHub actions afaik (https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/).

TristanBilot (approved):

PR #23, opened 2024-01-19
Benchmark

TristanBilot:
It seems that the first results of the benchmark are not satisfying. 

Multiple runs over 4 different experiments lead to around 50% slower execution of `mlx_gpu` compared to `pyg_cpu` on GCNConv and GATConv 🤔


francescofarina:
@TristanBilot That's quite interesting. But if I remember correctly, there are still some inefficiencies in convolutions in `mlx` (up to -70% wrt mps) - can it be due to that?

TristanBilot:
I don't think it has something to do with convolutions, as we're not using them here. I will check further if it's coming from scatter.

francescofarina:
Great! That's quite a good finding though, probably worth exposing to mlx.

TristanBilot:
I still didn't find the root cause of the performance bottleneck. I did some experiments with `scatter_add`, and there is not a huge difference between the `torch_scatter` and ours from mlx which runs on gpu. This may be a first clue. I will add comprehensive benchmarks of scatter operations on Monday.

francescofarina:
FWIW: I'm getting similar results also on my M3. Also interesting how mps is the slowest.

TristanBilot:
Nice, interesting indeed.
I just made [this issue](https://github.com/ml-explore/mlx/issues/506), Awni told me it could certainly optimized as the default scatter operations are not made for the situation where we have many indices and messages and few outputs nodes.

francescofarina:
@TristanBilot note that I believe `torch_scatter` doesn't run on `mps` - so it may be worth testing against (`torch.scatter`)[https://pytorch.org/docs/stable/generated/torch.scatter.html].

TristanBilot:
Ok I'll test that. Anyway, mps is not enough stable to be used in real GNN projects with PyG, so it will be just indicative. The real comparison is between MLX GPU and torch CPU.

francescofarina:
@TristanBilot can you sync this with the main branch and move the benchmarks into the new `benchmarks` folder? Then I think this can be safely merged and we can iterate later as the performance of mlx improves.

PR #24, opened 2024-01-19
Batch

Kheims:
## Proposed changes

This is a draft PR to allow discussion on the batching strategy.

The Batch class mimics the behaviour of pyg's one.

Usage : 
```

node_features1 = mx.array([[1,1,1,1],
            [1,1,1,1],
            [1,0,1,1],
            [0,0,1,1]])
edge_index1 = mx.array([[0,1,1,2,3],
                        [1,0,2,3,1]])
node_features2 = mx.array([[1,1,0,1],
            [1,0,1,1],
            [1,0,0,1]])
edge_index2 = mx.array([[0,1,2,3,1],
                        [1,0,1,2,2]])
g1 = GraphData(node_features=node_features1, edge_index=edge_index1)
g2 = GraphData(node_features=node_features2, edge_index=edge_index2)

batch = Batch.from_graph_list([g1,g2])
#Batch(node_features=[[1, 1, 1, 1], [1, 1, 1, 1], [1, 0, 1, 1], [0, 0, 1, 1], [1, 1, 0, 1], [1, 0, 1, 1], [1, 0, 0, 1]], edge_index=[[0, 1, 1, 2, 3, 4, 5, 6, 7, 5], [1, 0, 2, 3, 1, 5, 4, 5, 6, 6]],)

batch[0]
#GraphData(edge_index=[[0, 1, 1, 2, 3], [1, 0, 2, 3, 1]], node_features=[[1, 1, 1, 1], [1, 1, 1, 1], [1, 0, 1, 1], [0, 0, 1, 1]])

```
One of the points to take into consideration : 

- Specifically the ``__getitem__`` returns a freshly created GraphData object, shouldn't we aligned to pyg have a store to store these  ? 
- The from_graph_list for now only handle node_features and edge_index, we can add the handling of other attributes but what about user set attributes ? 

## Checklist


mlx_graphs/data/batch.py
```diff
@@ -8,35 +7,38 @@
...
-        list of :class:`~mlx_graphs.data.GraphData` 
+        list of :class:`~mlx_graphs.data.GraphData`
         """
-        
+
         if not isinstance(graphs, (list, tuple)):
             graphs = list(graphs)
 
         num_nodes = mx.array([0] + [graph.num_nodes() for graph in graphs])
+        cumsum = mx.cumsum(num_nodes)
```
TristanBilot, line 31: I replaced your num_nodes sum with cumsum. I don't know why you used that?

mlx_graphs/data/batch.py
```diff
@@ -8,35 +7,38 @@
...
-        global_edge_index = [graph.edge_index + num_nodes[i] for i,graph in enumerate(graphs)]
+        global_edge_index = [
+            graph.edge_index + cumsum[i] for i, graph in enumerate(graphs)
+        ]
         global_edge_index = mx.concatenate(global_edge_index, axis=1)
 
-        return cls(global_node_features, global_edge_index, num_nodes = num_nodes)
+        return cls(global_node_features, global_edge_index, cumsum=cumsum)
 
     def __getitem__(self, idx):
```
TristanBilot, line 43: Have to write tests for this function please

mlx_graphs/data/batch.py
```diff
@@ -52,33 +54,17 @@ def __getitem__(self, idx):
...
-        mask = (self.edge_index >= lower_last_bound) &\
-                (self.edge_index <=upper_bound)
-        
+        mask = (self.edge_index >= lower_last_bound) & (self.edge_index <= upper_bound)
+
         mask = mask[0] & mask[1]
 
         # NOTE : since boolean indexing isn't yet available we need to deduce the indices
-        indices = mx.array([i for i,e in enumerate(mask) if e])
+        indices = mx.array([i for i, e in enumerate(mask) if e])
```
TristanBilot, line 62: We'll have to find a better way to do that

mlx_graphs/data/batch.py
```diff
@@ -8,35 +7,38 @@
...
             graphs = list(graphs)
 
         num_nodes = mx.array([0] + [graph.num_nodes() for graph in graphs])
+        cumsum = mx.cumsum(num_nodes)
 
         global_node_features = [graph.node_features for graph in graphs]
         global_node_features = mx.concatenate(global_node_features, axis=0)
 
-        global_edge_index = [graph.edge_index + num_nodes[i] for i,graph in enumerate(graphs)]
+        global_edge_index = [
```
TristanBilot, line 36: The batching has to be done on the other arrays that constitute a GraphData, not only node features and edge index

mlx_graphs/data/batch.py
```diff
@@ -8,35 +7,38 @@
...
     """
-        Represents a batch data object describing a batch of graphs as one big (disconnected)
-        graph.
+    Represents a batch data object describing a batch of graphs as one big (disconnected)
+    graph.
 
     """
 
-    def __init__(self, node_features, edge_index, num_nodes) -> None:
+    def __init__(self, node_features, edge_index, cumsum) -> None:
```
TristanBilot, line 15: Missing batch_size arg

TristanBilot (changes requested):

mlx_graphs/data/batch.py
Kheims: It was a mistake indeed.

mlx_graphs/data/batch.py
Kheims: cls ? it's an internal function to construct an object out of the class the classmethod is part of, if you are referring to from_graph_list, the unit test, test it

mlx_graphs/data/batch.py
Kheims: Yes, wanted to have your opinions before pursuing elaborating the strategy for the other attributes that constitues a GraphData

mlx_graphs/data/batch.py
Kheims: What would batch_size do in this case ? isn't it directly inferred from the length of the graph list provided ? 

I'd have thinked that when creating a Dataset object there'd be a constructor that when a list of graphs is provided he would create n_batches, where n_batches = len(graphs) % batch_size ( so if we have a list of 10 graphs and specificy a batch_size of 3 we'd have 4 batches, 3 of them have each 3 graphs and the last only one )

mlx_graphs/data/batch.py
TristanBilot: No __getitem__ ^^

mlx_graphs/data/batch.py
TristanBilot: Yes you're right, sorry for the oversight, batch size will be located in a dataset or loader class!

francescofarina:
@Kheims @TristanBilot, a couple of points I think we should discuss here:
- In which cases one would initialize a `Batch` through the current `__init__` and not the `from_graph_list` method? I'd probably make what is currently `from_graph_list` the `Batch` initializers (meaning that I provide a `list[GraphData]` and I get a `Batch` object
- In general, I was wondering whether the PyG approach is the best one to follow here. Ideally, one could also extend the `GraphData` class to hold additional attributed that can be used to retrieve any batched graph that may constitute the main one. In that way we could just have a collator function `from_graph_list` taking as input a `list[GraphData]` and returning a `GraphData` (which now will be a disconnected one). Then we'd just have to implement the dataloaders which can use a graph collator to create batches. In the end a batched graph is just a graph.
A similar approach is actually used in [dgl](https://www.dgl.ai), where they have a `DGLGraph`, a [`batch`](https://docs.dgl.ai/_modules/dgl/batch.html) function and [`GraphDataLoader`](https://docs.dgl.ai/generated/dgl.dataloading.GraphDataLoader.html) (or a [`DataLoader`](https://docs.dgl.ai/generated/dgl.dataloading.DataLoader.html#dgl.dataloading.DataLoader)).

TristanBilot:
I like the way of batching graphs in dgl using `dgl.batch([g1, g2])`, maybe it's more straightforward than `from_graph_list()`? I see that dgl returns a new `Data` object after calling `batch()`, with new batch-related attributes. We have to think whether it's useful to have a dedicated `Batch` class or not.

If we want to follow dgl's paradigm, we have to create 2 classes for loading batches: `DataLoader` that will sample neighbors from a large graph and yield batches of neighbors, and `GraphDataLoader` that creates n batches of graphs and yields batches of graphs, for graph classification tasks.

francescofarina:
> I like the way of batching graphs in dgl using `dgl.batch([g1, g2])`, maybe it's more straightforward than `from_graph_list()`? 

Yes, that's probably ok. I believe most of the time the user won't have to be exposed to that part and the dataloader will take care of it if we move forward with the approach above.

> I see that dgl returns a new `Data` object after calling `batch()`, with new batch-related attributes. We have to think whether it's useful to have a dedicated `Batch` class or not.

I'd personally prefer to not have a dedicated `Batch` class for the reasons above. When you create a batch by collating multiple graphs, you end up with a new (disconnected) graph, but still a graph.
 
> If we want to follow dgl's paradigm, we have to create 2 classes for loading batches: `DataLoader` that will sample neighbors from a large graph and yield batches of neighbors, and `GraphDataLoader` that creates n batches of graphs and yields batches of graphs, for graph classification tasks.

I think that would go well also with the concerns we were discussing around loading data into memory, wdyt?

TristanBilot:
> I'd personally prefer to not have a dedicated `Batch` class for the reasons above. When you create a batch by collating multiple graphs, you end up with a new (disconnected) graph, but still a graph.

I agree, but we have to find a proper way to add the batch related attributes to the `GraphData` object. We have to kind of centralize the attributes that differ from a traditional `GraphData` and a batched `GraphData`.

TristanBilot:
> I think that would go well also with the concerns we were discussing around loading data into memory, wdyt?

Yes we can think of creating dedicated loaders for large graphs, which inherit from these loaders.

francescofarina:
> I agree, but we have to find a proper way to add the batch related attributes to the `GraphData` object. We have to kind of centralize the attributes that differ from a traditional `GraphData` and a batched `GraphData`.

Can't all `GraphData` have those attributes? For non-batched ones they'll just be `None`.

TristanBilot:
In the case we want to index a batch to get the n-th graph of a batch, we have to find back the original indices and nodes ith __getitem__(), and we have to run this only in batch mode, so we have to keep track clearly of which `GraphData` are batch or not.

TristanBilot:
Also, we may have some methods that we want to use only for batches and not for standalone graphs. Putting all together may not be the best way. However, we can say that a Batch inherits from `GraphData`, so we set all the graph-related attributes within the super() `GraphData` object.

Kheims:
I agree with @TristanBilot  comment here, it seems more compact to take the inheritance path and it also makes handling the exceptions ( for dimensions mismatches when creating the batched big graph and so on ) clearer for the user.

francescofarina:
I agree. Let's have it as a child of `GraphData`.

francescofarina:
@Kheims will you refactor and update this to include a `GraphDataBatch` (or similar name) object inheriting from `GraphData` and a `batch(graphs: list[GraphData]) -> GraphDataBatch` function? Just making sure we coordinate.

Kheims:
@francescofarina Affirmative for the cited changes, will refactor and add additional tests this thursday.
 I was a little overwhelmed these past two weeks, so sorry for the late replies, but from next week i'll have much more free time to dedicate to the project

francescofarina:
No strong preference here. Many graph-based libraries use the first one but the second option is also fine. We can also keep both as the second usually requires the first anyways and just exposes an additional interface.

TristanBilot:
Collate function from pyg: [here](https://github.com/pyg-team/pytorch_geometric/blob/fb31db1957a3341fc6a407aa23f7f98041691bd3/torch_geometric/data/collate.py#L149).

TristanBilot:
How they handle the [concatenation axis in PyG](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.data.Data.html#torch_geometric.data.Data.__cat_dim__).

Kheims:
Still need to work on the indexing to make it work dynamically on all found attributes that correspond to initial attributes of the GraphData objects of the list of graphs

explore.ipynb
francescofarina, line 1: Can you remove this file from the repo?

mlx_graphs/data/batch.py
```diff
@@ -0,0 +1,69 @@
...
+        # undo the increment induced by the batching
+        edge_index = self.edge_index[:, indices] - lower_last_bound
+
+        return GraphData(
+            node_features=node_features,
+            edge_index=edge_index,
+        )
+
+
+def batch(graphs: List[GraphData]) -> GraphDataBatch:
```
francescofarina, line 50: I'd probably pass the collate function as an argument here (defaulting to `collate`) - there may be cases where the user wants to customize it.

francescofarina (commented):
This is looking good - just need to make it work for all attributes of `GraphData`

mlx_graphs/data/batch.py
```diff
@@ -47,7 +47,7 @@ def __getitem__(self, idx):
         )
 
 
-def batch(graphs: List[GraphData]) -> GraphDataBatch:
+def batch(graphs: List[GraphData], collate_fn: Callable = collate) -> GraphDataBatch:
     """
     Constructs a :class:`mlx_graphs.batch.Batch` object from a
     list of :class:`~mlx_graphs.data.GraphData`
```
TristanBilot, line 58: Docs is missing for args

tests/batch/test_batch.py
```diff
@@ -0,0 +1,56 @@
...
+            [1, 0, 1, 1],
+            [0, 0, 1, 1],
+            [1, 1, 0, 1],
+            [1, 0, 1, 1],
+            [1, 0, 0, 1],
+        ]
+    )
+
+    expected_global_edge_index = mx.array(
+        [[0, 1, 1, 2, 3, 4, 5, 6, 7, 5], [1, 0, 2, 3, 1, 5, 6, 5, 6, 6]]
```
TristanBilot, line 30: Tests do not pass yet. This should be:
`[[0, 1, 1, 2, 3, 4, 5, 6, 7, 5], [1, 0, 2, 3, 1, 5, 4, 5, 6, 6]]`

docs/source/api/nn/index.rst
```diff
@@ -14,3 +14,6 @@ Neural Networks
 	GraphNetworkBlock
 	GATConv
 	GCNConv
+	global_add_pool
```
TristanBilot, line 17: Idk why pooling stuff is here. Pooling is already merged to main so it shouldn't appear as new code 🤔 any idea @francescofarina ?

TristanBilot (commented):
Nice work yet! 🚀 
Now it remains handling arbitrary attributes, and handle all edge cases in tests concerning all types and shapes of attributes, along with backward indexing (`batch[-1]`, `batch[-2]` etc), and out-of-range indexing (`batch[1000]` -> raise proper error with max batch size).
And what about slice indexing guys? Shall we implement `batch[:10]` yet? @Kheims @francescofarina

TristanBilot:
Nice work yet! 🚀 
Now it remains handling arbitrary attributes, and handle all edge cases in tests concerning all types and shapes of attributes, along with backward indexing (`batch[-1]`, `batch[-2]` etc), and out-of-range indexing (`batch[1000]` -> raise proper error with max batch size).
And what about slice indexing guys? Shall we implement `batch[:10]` yet? @Kheims @francescofarina

francescofarina:
@TristanBilot Yes having slice indexing would be great and fairly easy. 

For some reasons I can't pull the PR branch right now and contribute so adding the suggestion below @Kheims :

```python
class GraphDataBatch(GraphData):
    def __init__(self, node_features, edge_index, cumsum, num_graphs, **kwargs) -> None:
        super().__init__(node_features=node_features, edge_index=edge_index, **kwargs)
        self.cumsum = cumsum
        self.num_graphs = num_graphs

    def __getitem__(self, idx: Union[int, slice]) -> Union[GraphData, list[GraphData]]:
        if isinstance(idx, int):
            return self._get_graph(idx)
        elif isinstance(idx, slice):
            start, stop, step = idx.start or 0, idx.stop or sys.maxsize, idx.step or 1
            return [self._get_graph(i) for i in range(start, stop, step)]
        raise TypeError("GraphDataBatch indices should be int or slice.")

    def _get_graph(self, idx: int):
        lower_last_bound = self.cumsum[idx].item()
        upper_bound = self.cumsum[idx + 1].item()
        large_graph_dict = {
            attr: getattr(self, attr)
            for attr in self.to_dict()
            if attr not in {"cumsum", "num_graphs", "slices"}
        }
        single_graph_dict = {}
        for attr in large_graph_dict:
            if self.__inc__(attr):
                mask = (getattr(self, attr) >= lower_last_bound) & (
                    getattr(self, attr) <= upper_bound
                )
                mask = mask[0] & mask[1]
                indices = mx.array([i for i, e in enumerate(mask) if e])
                value = getattr(self, attr)[:, indices] - lower_last_bound
                single_graph_dict[attr] = value
            else:
                single_graph_dict[attr] = getattr(self, attr)[
                    lower_last_bound:upper_bound
                ]

        return GraphData(**single_graph_dict)

```

TristanBilot:
This operation is very expensive: `indices = mx.array([i for i, e in enumerate(mask) if e])`.
I opened an issue dedicated to that but the `nonzero()` function to get indices where condition is verified is not likely going to be implemented soon.

We can use a more efficient solution without masking by leveraging the number of nodes of batches, I'll try to do it.

francescofarina:
> This operation is very expensive: `indices = mx.array([i for i, e in enumerate(mask) if e])`.
> 
> I opened an issue (https://github.com/ml-explore/mlx/issues/568) dedicated to that but the `nonzero()` function to get indices where condition is verified is not likely going to be implemented soon.
> 
> 
> 
> We can use a more efficient solution without masking by leveraging the number of nodes of batches, I'll try to do it.

We can also consider temporarily using `np.nonzero`.

francescofarina:
Closing this as it's been moved to https://github.com/TristanBilot/mlx-graphs/pull/39

PR #25, opened 2024-01-22
WIP: Refactoring

francescofarina:
## Proposed changes
This PR addresses https://github.com/TristanBilot/mlx-graphs/issues/17.

As mentioned in https://github.com/TristanBilot/mlx-graphs/issues/17, the main open point is whether to stick with the `edge_index` having shape `[2, N]` (with `N` the number of edges) or to adopt a `[N, 2]` shape (which would be in line with how we define the shapes of features).

Now, the problem is that usually `N`>> 2 and since arrays are C-contiguous having `[2, N]` is more efficient if one needs to frequently extract sources and targets. 

I've run some benchmarks to compare some of the utils we have when storing indices as rows and as columns. For some operations there's a speedup but in `get_src_dst_features` (which basically just extracts the rows, if the shape of `edge_index` is `[2, N]`, or the columns if it's `[N,2]`) there's a major slowdown.

| operation | [N, 2] index | [2, N] index | [N, 2] speedup |
| --- | --- | --- | --- |
| get_src_dst_features | 5.62μs | 3.24μs | -73.06% |
| sort_edge_index | 7.17μs | 7.95μs | 9.82% |
| sort_edge_index_and_features | 7.93μs | 8.80μs | 9.82% |
| to_edge_index | 45.02μs | 45.12μs | 0.22% |
| to_sparse_adjacency_matrix | 48.54μs | 53.96μs | 10.05% |
| to_adjacency_matrix | 92.69μs | 97.25μs | 4.69% |
| is_undirected | 290.45μs | 258.29μs | -12.45% |

## Checklist

### DIFF
--- a/examples/graph_network_block.py
+++ b/examples/graph_network_block.py
@@ -9,12 +9,12 @@ class NodeModel(Module):
         self,
         node_features_dim: int,
         edge_features_dim: int,
-        global_features_dim: int,
+        graph_features_dim: int,
         output_dim: int,
     ):
         super().__init__()
         self.model = Linear(
-            input_dims=node_features_dim + edge_features_dim + global_features_dim,
+            input_dims=node_features_dim + edge_features_dim + graph_features_dim,
             output_dims=output_dim,
         )
 
@@ -23,9 +23,9 @@ class NodeModel(Module):
         edge_index: mx.array,
         node_features: mx.array,
         edge_features: mx.array,
-        global_features: mx.array,
+        graph_features: mx.array,
     ):
-        destination_nodes = edge_index[1]
+        destination_nodes = edge_index[:, 1]
         aggregated_edges = mx.zeros([node_features.shape[0], edge_features.shape[1]])
         for i in range(node_features.shape[0]):
             aggregated_edges[i] = mx.where(
@@ -37,8 +37,8 @@ class NodeModel(Module):
             [
                 node_features,
                 aggregated_edges,
-                mx.ones([node_features.shape[0], global_features.shape[0]])
-                * global_features,
+                mx.ones([node_features.shape[0], graph_features.shape[0]])
+                * graph_features,
             ],
             1,
         )
@@ -50,12 +50,12 @@ class EdgeModel(Module):
         self,
         edge_features_dim: int,
         node_features_dim: int,
-        global_features_dim: int,
+        graph_features_dim: int,
         output_dim: int,
     ):
         super().__init__()
         self.model = Linear(
-            input_dims=2 * node_features_dim + edge_features_dim + global_features_dim,
+            input_dims=2 * node_features_dim + edge_features_dim + graph_features_dim,
             output_dims=output_dim,
         )
 
@@ -64,17 +64,17 @@ class EdgeModel(Module):
         edge_index: mx.array,
         node_features: mx.array,
         edge_features: mx.array,
-        global_features: mx.array,
+        graph_features: mx.array,
     ):
-        source_nodes = edge_index[0]
-        destination_nodes = edge_index[1]
+        source_nodes = edge_index[:, 0]
+        destination_nodes = edge_index[:, 1]
         model_input = mx.concatenate(
             [
                 node_features[destination_nodes],
                 node_features[source_nodes],
                 edge_features,
-                mx.ones([edge_features.shape[0], global_features.shape[0]])
-                * global_features,
+                mx.ones([edge_features.shape[0], graph_features.shape[0]])
+                * graph_features,
             ],
             1,
         )
@@ -86,12 +86,12 @@ class GlobalModel(Module):
         self,
         edge_features_dim: int,
         node_features_dim: int,
-        global_features_dim: int,
+        graph_features_dim: int,
         output_dim: int,
     ):
         super().__init__()
         self.model = Linear(
-            input_dims=node_features_dim + edge_features_dim + global_features_dim,
+            input_dims=node_features_dim + edge_features_dim + graph_features_dim,
             output_dims=output_dim,
         )
 
@@ -100,12 +100,12 @@ class GlobalModel(Module):
         edge_index: mx.array,
         node_features: mx.array,
         edge_features: mx.array,
-        global_features: mx.array,
+        graph_features: mx.array,
     ):
         aggregated_edges = edge_features.mean(axis=0)
         aggregated_nodes = node_features.mean(axis=0)
         model_input = mx.concatenate(
-            [aggregated_nodes, aggregated_edges, global_features], 0
+            [aggregated_nodes, aggregated_edges, graph_features], 0
         )
         return self.model(model_input)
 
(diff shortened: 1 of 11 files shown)


francescofarina:
@TristanBilot what do you think?

TristanBilot:
It is interesting to think whether we use (2,N) or (N,2) `edge_index` shapes.
However, looking at your benchmark, it seems that there is a major slowdown if we use (N,2) for `get_src_dst_features()`.
This function is used in **every** message passing calls before computing messages, meaning that this function has to be the fastest possible. All other functions are much less important compared to `get_src_dst_features()`.

TristanBilot:
I also noticed that `get_src_dst_features()` is a major bottleneck in the overall message passsing forward call. It is the second slower operation after `scatter()`. As `get_src_dst_features()` is essentially a `gather` operation with many indices, I benchmarked `gather` in https://github.com/ml-explore/mlx/issues/518, between mlx and torch.
The results show that mlx+cpu and mlx+gpu are still much slower than torch+cpu.
If using a (N,2) is even more slow, I'm not sure it's good idea using it. However, we can think of using csr and csc formats for faster indexing based on rows and columns.

TristanBilot:
A very [interesting article](https://www.pyg.org/ns-newsarticle-accelerating-pyg-on-intel-cpus) on how PyG accelerates message-passing on CPU. They seem to use the COO format with the standard gather-scatter approach, not CSR.

francescofarina:
Yes, I guess the main reason lies in arrays being C-continuous. Let's stick with `[2, N]` for the `edge_index` default shape and I'll save the benchmarks in `benchmarks` so that we can re-test in the future if needed.
I'll revert some changes and finish the refactoring before making this ready for review.

francescofarina:
This should be ready now. 

Summary:
- all methods now take inputs in the order specified in https://github.com/TristanBilot/mlx-graphs/issues/17
- we stick to the `[2,N]` (COO) representation for the `edge_index` as it's more efficient
- updated examples and documentation

TristanBilot:
Nice, I will take a look tomorrow as it's quite a long PR :)

benchmarks/coo_vs_csr_edge_index/csr_functions.py
```diff
@@ -0,0 +1,88 @@
+import mlx.core as mx
+import numpy as np
+from typing import Optional
+
+
+def get_src_dst_features_CSR(
```
TristanBilot, line 6: AFAIK the CSR format is not the same thing as having a (N, 2) edge_index. Is there any reason why you named it CSR? I would say it's just the shape of the COO matrix that changes.

Example of CSR matrix:

```python
indptr = np.array([0, 2, 3, 6])
indices = np.array([0, 2, 2, 0, 1, 2])
data = np.array([1, 2, 3, 4, 5, 6])
csr_matrix((data, indices, indptr), shape=(3, 3)).toarray()
array([[1, 0, 2],
       [0, 0, 3],
       [4, 5, 6]])
```

examples/graph_network_block.py
```diff
@@ -64,17 +64,17 @@ def __call__(
         edge_index: mx.array,
         node_features: mx.array,
         edge_features: mx.array,
-        global_features: mx.array,
+        graph_features: mx.array,
     ):
-        source_nodes = edge_index[0]
-        destination_nodes = edge_index[1]
+        source_nodes = edge_index[:, 0]
+        destination_nodes = edge_index[:, 1]
```
TristanBilot, line 70: you still use the (N,2) format here

examples/graph_network_block.py
```diff
@@ -115,17 +115,18 @@ def __call__(
 F_E = 1  # number of edge features
 F_U = 2  # number of global features
 
-edge_index = mx.array([[0, 0, 1, 2, 3], [1, 2, 3, 3, 0]])
+edge_index = mx.array([[0, 0, 1, 2, 3], [1, 2, 3, 3, 0]]).transpose()
```
TristanBilot, line 118: remove `transpose()` here too

mlx_graphs/utils/transformations.py
```diff
@@ -92,7 +93,7 @@ def to_adjacency_matrix(
 
     Args:
         edge_index (mlx.core.array): a [2, num_edges] array representing the source and target nodes of each edge
```
TristanBilot, line 95: what do you think of providing a consistent doc description for all `edge_index` occurrences? We may do that as well for `node_features` and `edge_features`? We now have many different doc descriptions to describe the same thing.

tests/utils/test_sorting.py
```diff
@@ -23,8 +23,8 @@ def test_sort_edge_index_and_features():
     assert mx.array_equal(expected_edge_index, s), "Index sorting failes"
     assert mx.array_equal(expected_edge_features, f), "Features sorting failed"
     # test 2d edge_features
-    edge_features = mx.array([[5, 3, 2, 4, 1, 6], [2, 4, 5, 3, 6, 1]])
-    expected_edge_features = mx.array([[1, 2, 3, 4, 5, 6], [6, 5, 4, 3, 2, 1]])
+    edge_features = mx.array([[5, 3, 2, 4, 1, 6], [2, 4, 5, 3, 6, 1]]).T
+    expected_edge_features = mx.array([[1, 2, 3, 4, 5, 6], [6, 5, 4, 3, 2, 1]]).T
```
TristanBilot, line 27: Is the transpose operation expected here? (just to be sure you didn't miss one)

mlx_graphs/utils/scatter.py
```diff
@@ -1,14 +1,15 @@
...
 
+ScatterAggregations = Literal["add", "max", "softmax"]
+
 
 def scatter(
     values: mx.array,
     index: mx.array,
     out_size: Optional[int] = None,
-    aggr: Literal["add", "max", "softmax"] = "add",
+    aggr: ScatterAggregations = "add",
```
TristanBilot, line 12: 👏

TristanBilot (changes requested):

benchmarks/coo_vs_csr_edge_index/csr_functions.py
francescofarina: Yes, you're right 💯  - it's a leftover of some other initial tests and I carried it over without paying attention. I'll rename them as `column` and `row`.

examples/graph_network_block.py
francescofarina: good catch!

tests/utils/test_sorting.py
francescofarina: This is actually correct as it is. `edge_features` is expected to have shape `[num_edges, num_edge_features]` and in this case there are 6 edges (and 2 features per edge) - it was just more convenient to write it as a row and transpose it.

mlx_graphs/utils/transformations.py
francescofarina: I actually thought about it as well but I'm not sure where to put it at this stage as the documentation is not published. Let me add it to the README and to the first page of the docs, what do you think @TristanBilot ?

TristanBilot: Ok we can do it later, we just have to remind that to have consistent docs in all methods, and yeah go for the README if you want!

TristanBilot (approved):

mlx_graphs/utils/transformations.py
francescofarina: I added it to the README and the docs. Let's decide later whether to remove the extensive descriptions from the individual classes/methods docs.

PR #29, opened 2024-01-24
Update benchmarks deps

TristanBilot:
## Proposed changes
Just updating the benchmarks deps.

## Checklist


francescofarina (approved):
🚢

PR #34, opened 2024-01-25
Pooling

TristanBilot:
## Proposed changes

This PR introduces Add, Max and Mean pooling operations + `scatter_mean` + `degree`.
Issue: (https://github.com/TristanBilot/mlx-graphs/issues/31)

## Checklist


tests/utils/test_scatter.py
```diff
@@ -1,6 +1,7 @@
 import mlx.core as mx
+import pytest
 
-from mlx_graphs.utils import scatter
+from mlx_graphs.utils import scatter, degree
 
 
 def test_scatter():
```
francescofarina, line 7: You can put `test_scatter` and `test_scatter_2D` together and parametrize the test

francescofarina (changes requested):
Just two minor things. 

QQ: why are you using r-strings as docstrings?

francescofarina:
Can you also add the pooling layers and the `degree` op to the documentation?

TristanBilot:
> QQ: why are you using r-strings as docstrings?

The scatter operations don't appear in detail on the html docs so I thought it may be due to that, do you have any idea? I also see other open source projects doing that, I'm not sure to get why.

TristanBilot:
> Can you also add the pooling layers and the `degree` op to the documentation?

Oh so I have to create rst files by myself? Never used sphinx before.

francescofarina:
> > Can you also add the pooling layers and the `degree` op to the documentation?
> 
> Oh so I have to create rst files by myself? Never used sphinx before.

Not necessarily the whole files. You can add `scatter_mean` and `degree` [here](https://github.com/TristanBilot/mlx-graphs/blob/main/docs/source/api/utils/scatter.rst) and the polling ops [here](https://github.com/TristanBilot/mlx-graphs/blob/main/docs/source/api/nn/index.rst) since they're imported in `mlx_graphs/nn/__init__.py`

francescofarina:
> > QQ: why are you using r-strings as docstrings?
> 
> The scatter operations don't appear in detail on the html docs so I thought it may be due to that, do you have any idea? I also see other open source projects doing that, I'm not sure to get why.

This should be solved once you modify the docs files.

tests/utils/test_scatter.py
```diff
@@ -1,39 +1,97 @@
...
+from mlx_graphs.utils import scatter, degree
 
 
-def test_scatter():
-    src = mx.array([1.0, 1.0, 1.0, 1.0])
-    index = mx.array([0, 0, 1, 2])
-    num_nodes = src.shape[0]
-
-    y_hat1 = scatter(src, index, aggr="softmax")
+@pytest.mark.parametrize(
```
TristanBilot, line 7: The thing with parametrized tests is that if a test fails, it doesn't show which parameters failed.

francescofarina: You should be able to see it by running pytest in verbose mode (`pytest -v`)

francescofarina (approved):
Amazing! 🚢 

Minor thing, you may want to replace `mx.array` with `mlx.core.array` in the docstrings to have the object show correctly in the documentation.

TristanBilot:
> Minor thing, you may want to replace `mx.array` with `mlx.core.array` in the docstrings to have the object show correctly in the documentation.

Nice catch, done.

PR #36, opened 2024-01-25
Update and refactor docs

francescofarina:
## Proposed changes

Fixing https://github.com/TristanBilot/mlx-graphs/issues/35 and refactoring docs. 
I added the `sphinx-autodoc-typehints` plugin so that types in docstrings are directly inferred from the function signatures. This is in line with PEP484.

## Checklist


francescofarina:
NB: from now on we won't write explicit types in the docstrings as we already specify them in the signatures.

TristanBilot (approved):
🚢

PR #39, opened 2024-01-28
Batch2

TristanBilot:
This is a duplicated branch of @Kheims' work for batching as the first branch was broken.
I integrated batch for arbitrary array attributes. `__cat_dim__` and `__incr__` have been implemented and tested. Slicing and indexing with negative numbers is also done. 

I will add docs, edge case errors and more tests tomorrow.


mlx_graphs/data/batch.py
```diff
@@ -0,0 +1,107 @@
+from typing import List, Union
+import mlx.core as mx
+
+from mlx_graphs.data.data import GraphData
+from mlx_graphs.data.collate import collate
+
+
+class GraphDataBatch(GraphData):
+    def __init__(self, graphs, **kwargs) -> None:
```
francescofarina, line 9:
```suggestion
    def __init__(self, graphs: list[GraphData], **kwargs) -> None:
```

mlx_graphs/data/batch.py
```diff
@@ -0,0 +1,107 @@
+from typing import List, Union
```
francescofarina, line 1: Importing `List` is not needed anymore - we can directly use `list` (see PEP484 or [here](https://docs.python.org/3/library/typing.html)).

mlx_graphs/data/collate.py
```diff
@@ -0,0 +1,51 @@
+from typing import List
```
francescofarina, line 1: As above, no need for `List`

mlx_graphs/data/batch.py
```diff
@@ -0,0 +1,107 @@
...
+
+            # If __inc__ is True, we get back the original values by subtracting
+            # the cumulative sum at the batch index
+            if self.to_dict()[f"_inc_{attr}"]:
+                original_value -= self._cumsum[idx]
+
+            # NOTE: `self._cumsum` is the cumsum of number of nodes per graph. For custom features
+            # with __inc__, each feature should have its own cumsum
+
+            single_graph_dict[attr] = original_value
```
francescofarina, line 73: this is very neat!

mlx_graphs/data/collate.py
```diff
@@ -0,0 +1,51 @@
...
+    apart from `edge_index` which is concatenated along dim 1 and
+    nodes indices are incremented
+
+    Args:
+        graph_list (List[GraphData]): the list of GraphData objects to collate
+
+    Returns:
+        dict: dict containing all the attributes of the unified and disconnected big graph as
+        well as the slices corresponding to the portion of node for each graph from the provided
+        list of graphs.
```
francescofarina, line 20: No need for type specs in docstrings as they're now inferred from the signature

```suggestion
    Args:
        graph_list: the list of GraphData objects to collate

    Returns:
        Dict containing all the attributes of the unified and disconnected big graph as
        well as the slices corresponding to the portion of node for each graph from the provided
        list of graphs.
```

mlx_graphs/data/collate.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    batch_attr_dict = {}
+
+    for attr in graph_list[0].to_dict():
```
francescofarina, line 24: This is very much an edge case, but what if some graphs in the list have additional attributes? Should they be concatenated in some way? Ignored? Should we throw an error?

francescofarina (changes requested):
Just some minor initial comments! I'll re-review after docs and full tests are implemented

mlx_graphs/data/collate.py
TristanBilot: Ahha that's exactly what I planed to work on today. I think we should either ignore other attributes, or raise an error. 
Just ignoring might by tricky for the user if he's looking for those values in the final batch, and raising might be a bit too strong, but I don't see why some GrahData would have additional attributes.

francescofarina: Yeah, I agree that's something that shouldn't happen, at least for homogeneous graphs - it could if they're heterogeneous though, but we can leave that for later!

I think in this case we should raise an error as having extra attributes is something we don't expect, or we can raise a warning saying that extra attributes will be ignored, wdyt?

TristanBilot: Okay let's go for the raising option!

francescofarina: The only thing I'd be slightly worried about is the overhead we'll incur in checking all the graphs in the batch - but it should be a fairly cheap operation.

docs/source/api/data/batch.rst
```diff
@@ -0,0 +1,12 @@
...
+Batch Objects
+=============
+
+.. currentmodule:: mlx_graphs.data.batch
+
+
+.. autosummary::
+	:toctree: _autosummary
+
+	GraphDataBatch
```
TristanBilot, line 12: @francescofarina do you know how to consider `__incr__` and `__cat_dim__` methods in the docs?

francescofarina: We'll likely have to write a custom template (right now I wrote only this docs/source/_templates/custom-class-template.rst for the nn module to include the `__call__`). Why do you want to include them in the doc?

TristanBilot: It's important to have these in the docs as the user can override them for custom batching.

francescofarina: Ok, then a custom template is the way to go.

mlx_graphs/data/batch.py
```diff
@@ -0,0 +1,167 @@
+from typing import Union
+import mlx.core as mx
+
+from mlx_graphs.data.data import GraphData
+from mlx_graphs.data.collate import collate
+
+
+class GraphDataBatch(GraphData):
+    """Concatenates multiple `GraphData` into a single unified `GraphBatch`
```
francescofarina, line 9: `GraphBatch` -> `GraphDataBatch`

mlx_graphs/data/batch.py
```diff
@@ -0,0 +1,167 @@
...
+                node_features=array([[1], [1], [1]], dtype=float32)),
+            GraphData(
+                edge_index=array([[3, 3, 3], [4, 4, 4]], dtype=int32),
+                node_features=array([[2], [2], [2]], dtype=float32))
+        ]
+    """
+
+    def __init__(self, graphs: list[GraphData], **kwargs) -> None:
+        if not isinstance(graphs, (list, tuple)):
+            graphs = list(graphs)
```
francescofarina, line 66: I think we can safely remove this as it's clearly documented that the argument needs to be a list of graph. We can eventually add the decorator `@validate_list_of_graph_data`

mlx_graphs/data/batch.py
```diff
@@ -0,0 +1,167 @@
...
+    list of :class:`~mlx_graphs.data.GraphData`
+    """
+    return GraphDataBatch(graphs)
+
+
+def unbatch(batch: GraphDataBatch) -> list[GraphData]:
+    """Reconstructs a list of :class:`~mlx_graphs.data.GraphData`
+    objects from the :class:`~mlx_graphs.data.GraphDataBatch` object.
+    """
+    return [batch[idx] for idx in range(batch.num_graphs)]
```
francescofarina, line 167: update docs

docs/source/api/data/batch.rst
TristanBilot: ok we'll handle that in another PR

mlx_graphs/data/collate.py
```diff
@@ -0,0 +1,56 @@
...
+    apart from `edge_index` which is concatenated along dim 1.
+    Each graph remains independent in the final graph by incrementing
+    the indices in `edge_index` based on the cumsum of previous number
+    of nodes per graph.
+
+    Args:
+        graph_list: List of `GraphData` objects to collate
+
+    Returns:
+        dict: dict containing all the attributes of the unified and disconnected big graph as
```
francescofarina, line 22:
```suggestion
        Dict containing all the attributes of the unified and disconnected big graph as
```

mlx_graphs/data/data.py
```diff
@@ -47,19 +47,61 @@ def __init__(
...
+
+        By default, all attribute names containing "index" will be concatenated on axis 1,
+        e.g. `edge_index`. Other attributes are concatenated on axis 0, e.g. node features.
+
+        Args:
+            key: Name of the attribute on which change the default concatenation dimension
+                while using batching.
+
+        Returns:
+            int: The dimension where concatenation will happen when batching.
```
francescofarina, line 82:
```suggestion
            The dimension where concatenation will happen when batching.
```

mlx_graphs/data/data.py
```diff
@@ -47,19 +47,61 @@ def __init__(
...
+        By default, all attribute names containing "index" will be incremented to avoid
+        duplicate nodes in the index, e.g. `edge_index`.
+        Other attributes are cnot incremented and keep their original values, e.g. node features.
+
+        Args:
+            key: Name of the attribute on which change the default incrementation behavior
+                while using batching.
+
+        Returns:
+            bool: Whether to use incrementation of values for a given attribute.
```
francescofarina, line 103:
```suggestion
            Whether to use incrementation of values for a given attribute.
```

tests/nn/test_message_passing.py
```diff
@@ -59,9 +59,9 @@ def test_sum_aggregation():
     y3 = mx.array([[2, 4], [5, 6], [0, 0]])
     y_hat3 = mpnn(x, edge_index, edge_weight=None)
 
-    edge_index = mx.array([[0, 100, 0], [0, 1, 100]])
-    y4 = mx.array([[1, 2], [0, 0], [0, 0]])
-    y_hat4 = mpnn(x, edge_index, edge_weight=None)
+    # edge_index = mx.array([[0, 100, 0], [0, 1, 100]])
+    # y4 = mx.array([[1, 2], [0, 0], [0, 0]])
+    # y_hat4 = mpnn(x, edge_index, edge_weight=None)
```
francescofarina, line 64: What error do you get here with `pytest`? Looks like it works ok for me

francescofarina (changes requested):

tests/nn/test_message_passing.py
francescofarina: No, now failing for me as well

mlx_graphs/data/batch.py
TristanBilot: Ok, `collate()` already handles `@validate_list_of_graph_data` so no need for an additional call.

mlx_graphs/data/batch.py
TristanBilot: Do we need `:class:~mlx_graphs.data.GraphData` here or we still use `GraphData` like everywhere else.

francescofarina: I don't think we need it, especially if you add `args` and `returns`

tests/nn/test_message_passing.py
TristanBilot: Yes it's a strange one

francescofarina:
@TristanBilot I added the docs template here https://github.com/TristanBilot/mlx-graphs/pull/41  - it will be merged into this branch

TristanBilot:
Ok nice, you can merge it into this one when you're done.

TristanBilot:
I just implemented `batch_indices`, which is similar to `batch` in PyG.

TristanBilot:
I got few errors with the docs:
`mlx_graphs.data.batch.GraphDataBatch.rst: WARNING: document isn't included in any toctree`
`mlx_graphs.data.batch.GraphDataBatch, use :no-index: for one of them`

francescofarina:
> I got few errors with the docs: `mlx_graphs.data.batch.GraphDataBatch.rst: WARNING: document isn't included in any toctree` `mlx_graphs.data.batch.GraphDataBatch, use :no-index: for one of them`

Try deleting the `docs/build` and removing the generated `_autosummary*` subfolders (in source) before compiling the docs, you may have some outdated files. It works well for me

mlx_graphs/data/data.py
```diff
@@ -47,19 +47,62 @@ def __init__(
...
 
         Returns:
-            dict: A dictionary representation of the Data object.
+            A dictionary with all attributes of the `GraphData` object.
         """
         return {k: v for k, v in self.__dict__.items() if v is not None}
+
+    @property
+    def num_nodes(self) -> int:
+        return self.node_features.shape[0]
```
francescofarina, line 67: `self.node_features` may be None

```suggestion
    @property
    def num_nodes(self) -> Optional[int]:
        if self.node_features:
            return self.node_features.shape[0]
        return None
```

mlx_graphs/data/data.py
```diff
@@ -47,19 +47,62 @@ def __init__(
...
+
+        Args:
+            key: Name of the attribute on which change the default incrementation behavior
+                while using batching.
+
+        Returns:
+            Incrementing value for the given attribute or None.
+        """
+        if "index" in key:
+            return len(self.node_features)
```
francescofarina, line 107: `self.node_features` may be None

francescofarina (changes requested):
Just two extra comments in `data.py` as `self.node_features` may be None

mlx_graphs/data/data.py
francescofarina:
```suggestion
            return len(self.node_features) if self.node_features else None
```

mlx_graphs/data/data.py
TristanBilot: The None node_features is unexpected behavior for the moment, we didn't handle this case in messsage passing yet. 
I think I will have to do other modifications in the project to handle this case properly. 
Are node features also optional in PyG?

francescofarina: Yes they are - see [here](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/data/data.html#Data). In general all attributes are optional so we'll have to handle that

francescofarina: We can also do that in another PR btw.

TristanBilot:
I think I'm done with this PR, I'll add more handling with None node features later in another PR

francescofarina:
Cool, merging now

PR #44, opened 2024-01-29
Fix and test graphs validator

francescofarina:
## Proposed changes

Fix `validate_list_of_graph_data` to check argument is a `list[GraphData]` + tests + docs

## Checklist

### DIFF
--- a/mlx_graphs/utils/validators.py
+++ b/mlx_graphs/utils/validators.py
@@ -65,13 +65,18 @@ def validate_list_of_graph_data(func):
 
     @functools.wraps(func)
     def wrapper(graph_list: list[GraphData], *args, **kwargs):
-        expected_attr = set(graph_list[0].to_dict())
+        if not isinstance(graph_list, list):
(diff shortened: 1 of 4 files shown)


mlx_graphs/utils/validators.py
```diff
@@ -65,13 +65,18 @@ def validate_list_of_graph_data(func):
...
+
         for i, graph in enumerate(graph_list):
             if not isinstance(graph, GraphData):
                 raise ValueError(
+                    "Expected list of GraphData. "
                     f"Graph at index {i} in the batch is not of type `GraphData`."
                 )
 
+        expected_attr = set(graph_list[0].to_dict())
+        for i, graph in enumerate(graph_list):
```
TristanBilot, line 79: Why doing the same loop twice?

francescofarina: because if the elements are not `GraphData`, `graph_list[0]` don't have the `to_dict` method. Anyways, you're right it's not efficient so I changed how `expected_attr` is handled

Issue #46, opened 2024-01-29
num_hops()


TristanBilot:
Easy to implement and could be useful: [PyG docs](https://pytorch-geometric.readthedocs.io/en/latest/modules/utils.html#torch_geometric.utils.get_num_hops).

PR #51, opened 2024-01-30
Get num hops

Kheims:
## Proposed changes

Implementation of [#46 ](https://github.com/TristanBilot/mlx-graphs/issues/46).

I have used misc as the module name, do you have a better name in mind or maybe suggest to relocate it in other existing modules ?

## Checklist


tests/utils/test_misc.py
```diff
@@ -0,0 +1,22 @@
...
+            self.conv1 = GCNConv(4, 16)
+            self.conv2 = GCNConv(16, 16)
+            self.lin = nn.Linear(16, 2)
+
+        def __call__(self, edge_index: mx.array, node_features: mx.array) -> mx.array:
+            x = nn.relu(self.conv1(node_features, edge_index))
+            x = self.conv2(node_features, edge_index)
+            return self.lin(x)
+
+    assert get_num_hops(GCN()) == 2, "get_num_hops failed"
```
TristanBilot, line 22: Can you add a test with a 0-hop MPNN?

PR #52, opened 2024-01-30
SAGEConv

TristanBilot:
## Checklist

### DIFF
--- a/mlx_graphs/data/__init__.py
+++ b/mlx_graphs/data/__init__.py
@@ -1,0 +1,6 @@
+# from .data import GraphData # noqa
+# from .batch import (GraphDataBatch, batch, unbatch) # noqa
+# from .collate import collate # noqa
+
+# TODO: when uncommenting, this generates a circular import error, need tovfix that
+# to import GraphData with from mlx_graphs.data import GraphData
--- a/mlx_graphs/nn/conv/sage_conv.py
+++ b/mlx_graphs/nn/conv/sage_conv.py
@@ -1,0 +1,113 @@
+from typing import Optional
+
+import mlx.core as mx
+
+from mlx_graphs.nn.linear import Linear
+from mlx_graphs.nn.message_passing import MessagePassing
+
+
+class SAGEConv(MessagePassing):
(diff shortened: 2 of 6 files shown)


mlx_graphs/data/__init__.py
```diff
@@ -0,0 +1,6 @@
+# from .data import GraphData # noqa
+# from .batch import (GraphDataBatch, batch, unbatch) # noqa
+# from .collate import collate # noqa
+
+# TODO: when uncommenting, this generates a circular import error, need tovfix that
+# to import GraphData with from mlx_graphs.data import GraphData
```
francescofarina, line 6: I don't see why that's happening. I just gave it a try and I don't have a circular import error when doing `from mlx_graphs.data import GraphData`.

mlx_graphs/nn/conv/sage_conv.py
```diff
@@ -0,0 +1,113 @@
...
+            neigh_features = self.propagate(
+                edge_index=edge_index, node_features=node_features, **message_kwargs
+            )
+            neigh_features = self.neigh_proj(neigh_features)
+
+        out_features = self.self_proj(node_features) + neigh_features
+
+        return out_features
+
+    def message(
```
francescofarina, line 103: I'm wondering whether we should make this method private - but I guess it's ok as it is

francescofarina (approved):
All good - just the think on circular imports

mlx_graphs/nn/conv/sage_conv.py
TristanBilot: hmm, we have to think about that. It all depends more on the UX than on the good practices here

mlx_graphs/data/__init__.py
francescofarina: FWIW - all tests run well on my end

TristanBilot: Did you change the occurrences: `from mlx_graphs.data.data import GraphData` to `from mlx_graphs.data import GraphData`?

francescofarina: I did, both in `test_data` and `test_batch`

PR #53, opened 2024-01-30
Get num hops

Kheims:
## Proposed changes

Implementation of https://github.com/TristanBilot/mlx-graphs/issues/46.
I have used misc as the module name, do you have a better name in mind or maybe suggest to relocate it in other existing modules ? I agree it is related to the models should we have a utils/ or model_info/ or something like that located in nn ?

## Checklist


mlx_graphs/utils/misc.py
```diff
@@ -0,0 +1,36 @@
+import mlx.nn as nn
+
+
+def get_num_hops(model: nn.Module) -> int:
+    """
+    Returns the number of hops the model is aggregating information
+    from.
```
francescofarina, line 7: Specify that this works only for networks based on `MessagePassing`

docs/source/api/utils/misc.rst
```diff
@@ -0,0 +1,14 @@
...
+
+Misc
+=======
+
+
+.. currentmodule:: mlx_graphs.utils.misc
+
+
+.. autosummary::
+	:toctree: _autosummary_sorting
```
francescofarina, line 12: `_autosummary_misc`

francescofarina (changes requested):
Let's keep it in `mlx_graphs/utils` but let's find a better name than `misc`. Perhaps it should stay in `topology.py`?

mlx_graphs/utils/misc.py
francescofarina, line 1: rm file?

Kheims:
> Let's keep it in `mlx_graphs/utils` but let's find a better name than `misc`. Perhaps it should stay in `topology.py`?

Although topology seems graph's structure related, I think it's indeed better to put it there.

francescofarina (approved):
Looks good - please remove `misc.py` before merging

mlx_graphs/utils/misc.py
Kheims: Yes just did

mlx_graphs/utils/topology.py
```diff
@@ -62,3 +63,38 @@ def is_directed(edge_index: mx.array, edge_features: Optional[mx.array] = None)
...
+        ...         self.conv1 = GCNConv(4, 16)
+        ...         self.conv2 = GCNConv(16, 16)
+        ...         self.lin = nn.linear(16, 2)
+        ...
+        ...     def __call__(self, edge_index: mx.array, node_features: mx.array) -> mx.array:
+        ...         x = nn.relu(self.conv1(node_features, edge_index))
+        ...         x = self.conv2(node_features, edge_index)
+        ...         return self.lin(x)
+        >>> get_num_hops(GNN())
+        2
```
francescofarina, line 92: @Kheims can you please make this a code-block?

PR #54, opened 2024-01-30
GINConv

TristanBilot:
This PR introduces:

* GINConv
* Modifiable `aggr` in all layers
* Handle `edge_weights` by default in all MPNN layers
* `edge_weights` implem' for GCN

## Checklist

### DIFF
--- a/mlx_graphs/nn/conv/gin_conv.py
+++ b/mlx_graphs/nn/conv/gin_conv.py
@@ -1,0 +1,99 @@
+from typing import Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from mlx_graphs.nn.message_passing import MessagePassing
+
+
+class GINConv(MessagePassing):
+    r"""Graph Isomorphism Network convolution layer from `"How Powerful are
+    Graph Neural Networks?" <https://arxiv.org/abs/1810.00826>`_ paper.
+
+    .. math::
+        \mathbf{h}_i = \text{MLP} \left( (1 + \epsilon) \cdot
+        \mathbf{x}_i + \sum_{j \in \mathcal{N}(i)} \mathbf{x}_j \right)
+
+    where :math:`\mathbf{x}_i` and  :math:`\mathbf{h}_i` represent the input features
+    and output embeddings of node :math:`i`, respectively. :math:`\text{MLP}` denotes
+    a custom neural network provided by the user and :math:`\epsilon` is an epsilon
+    value either fixed or learned.
+
+    Args:
+        mlp: Callable :class:`mlx.core.nn.Module` applied on the final node embeddings
+        eps (optional): Initial value of the :math:`\epsilon` term. Default: ``0``
+        learn_eps (optional): Whether to learn :math:`\epsilon` or not. Default ``False``
+
+    Example:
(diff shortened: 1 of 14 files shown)


mlx_graphs/nn/conv/gin_conv.py
```diff
@@ -0,0 +1,99 @@
...
+        array([[-0.536501, 0.154826, 0.745569, ..., 0.31547, -0.0962588, -0.108504],
+            [-0.415889, -0.0498145, 0.597379, ..., 0.194553, -0.251498, -0.207561],
+            [-0.119966, -0.0159533, 0.276559, ..., 0.0258303, -0.194533, -0.15515],
+            [-0.21477, -0.169684, 0.485867, ..., 0.0194768, -0.145761, -0.139433],
+            [-0.133289, -0.0279559, 0.358095, ..., -0.0443346, -0.11571, -0.114396]], dtype=float32)
+    """
+
+    def __init__(
+        self,
+        mlp: nn.Module,
```
francescofarina, line 60: can we rename it to `model` or something similar?

mlx_graphs/nn/conv/gin_conv.py
```diff
@@ -0,0 +1,99 @@
...
+        \mathbf{x}_i + \sum_{j \in \mathcal{N}(i)} \mathbf{x}_j \right)
+
+    where :math:`\mathbf{x}_i` and  :math:`\mathbf{h}_i` represent the input features
+    and output embeddings of node :math:`i`, respectively. :math:`\text{MLP}` denotes
+    a custom neural network provided by the user and :math:`\epsilon` is an epsilon
+    value either fixed or learned.
+
+    Args:
+        mlp: Callable :class:`mlx.core.nn.Module` applied on the final node embeddings
+        eps (optional): Initial value of the :math:`\epsilon` term. Default: ``0``
```
francescofarina, line 24: Please remove all the `(optional)` from the docstrings. Types are inferred from the signature and if a value is provided, like `(optional)`, the actual type is overwritten in the docs.

francescofarina (changes requested):
Looking good - just two comments to address before merging.

mlx_graphs/nn/conv/gin_conv.py
TristanBilot: `mlp` is how it's called in the original paper, `model`  is too general. Do you prefer `nn` or `neural_net`?

TristanBilot: Or even `encoder`

mlx_graphs/nn/conv/gin_conv.py
TristanBilot: You're right

mlx_graphs/nn/conv/gin_conv.py
TristanBilot: The MLP is included in the equation, so it's quite clear where it's used when looking at the docs of the layer

francescofarina: ok, let's leave it as it is

francescofarina (approved):

PR #55, opened 2024-01-30
Github pages docs

TristanBilot:


TristanBilot:
I do some tests on this branch for GH pages. It seems that it didn't create the `gh-pages` branch, and [the page](https://tristanbilot.github.io/mlx-graphs/) does not render the links right. Any idea @francescofarina ?

francescofarina:
Have you set up the branch from which it should read as in https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site ?

Also I think we may have to do this step manually if there are issues on setting up the env for mlx in the action.

Basically to do I manually we need to build the docs locally (`make html`) and then force add the `build/html` directory (which is gitignored by default) `git add -f build/html` and then push the changes to the `gh-pages` branch.

francescofarina:
> Have you set up the branch from which it should read as in https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site ?

I assume you did it, so just checking - you have to point it to where the `index.html` is located, otherwise it defaults to the `README.md` in the root

TristanBilot:
Here is the current error: https://github.com/TristanBilot/mlx-graphs/actions/runs/7717444545/job/21036565423

TristanBilot:
You can easily try by yourself by pushing to this branch and check in `Actions`.

TristanBilot:
> I assume you did it, so just checking - you have to point it to where the `index.html` is located, otherwise it defaults to the `README.md` in the root

Regarding this, I need to specify the branch created by the action after the build which will be dedicated to the build files. Now this branch is not created as the build doesn't work yet. Once built, we can specify either root/ or docs/ depending on where the build files are located.

francescofarina:
I'm trying locally first to build the docs without `mlx-graphs` installed. I'm working on the `source/conf.py` where we need to mock the package modules at the beginning of the file as

```python
from unittest.mock import MagicMock


class Mock(MagicMock):
    @classmethod
    def __getattr__(cls, name):
        return MagicMock()


# Mock the modules you want to exclude from import during documentation build
MOCK_MODULES = [
    "mlx_graphs",
    "data",
    # "nn",
    # "utils",
]
sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)

```
It works for `data`, but when I add `nn` it breaks

TristanBilot:
Are we supposed to mock mlx_graphs modules?
Idk why the modules are not found as we run the same commands as we do locally.

francescofarina:
I'm trying to figure out what's going on. If you uninstall `mlx-graphs` from your local dev environment the docs won't compile as in the GitHub action.

TristanBilot:
If I do:
```
pip uninstall mlx_graphs
make clean
make html
```
It still works

francescofarina:
I need to delete all the `_autosummary*` folders to make it work but that shouldn't be a problem in the action.

francescofarina:
@TristanBilot Sphinx build is now working correctly. I'm not sure what's the aim of ghp-import as the last step

francescofarina:
We may want to use something like [this](https://github.com/peaceiris/actions-gh-pages) as the last step, with

```
- name: Deploy to external repos
      uses: peaceiris/actions-gh-pages@v3
      with:
        deploy_key: ${{ secrets.GITHUB_TOKEN }}
        publish_dir: docs/build/html
```

francescofarina:
This will publish the new documentation and push it to the `gh-pages` branch on every push to main.

@TristanBilot please make sure to update `Settings -> Pages -> Build and deployment` to `Deploy from a branch` with the branch selected to be `gh-pages` from `root`

Screenshot 2024-01-31 at 11 16 34

I have it working on my fork [here](https://francescofarina.github.io/mlx-graphs/)

TristanBilot:
Amazing!

Issue #64, opened 2024-02-02
Add tests to CI/CD


francescofarina:
macOS runners are now available for OS projects: https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

PR #65, opened 2024-02-02
Add tests to CI

francescofarina:
## Proposed changes

Change to run tests for PRs #64.

### DIFF
--- a/.github/workflows/pull_request_workflow.yml
+++ b/.github/workflows/pull_request_workflow.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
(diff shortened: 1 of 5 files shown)


francescofarina:
Using this to test the workflow

francescofarina:
For some reasons some operations performed in the tests break the CI, e.g. line 56 of `collate.py`.

TristanBilot:
We can really run tests on `macos-14`? 😳

francescofarina:
Yes ;) See [here](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/).

Everything is installed correctly including mlx but some tests seem to break the process. I know `test_batch` and `test_message_passing` cause the process to break but testing the utils seems to work fine.
I can't reproduce it locally so it may as well be an issue with the new runners.

TristanBilot:
hmm, so what do you recommend to do?

francescofarina:
Let me explore it more and also let's wait a few days as macOS runners have just been released and there may be issues to be fixed.

francescofarina:
Still no luck with this - and it's hard to debug as it can't be reproduced locally. I suggest we leave this open as a draft and go back to it in a few days. Meanwhile let's be diligent with local testing.

PR #67, opened 2024-02-02
isort

TristanBilot:
## Proposed changes

Introduces isort to sort imports.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,8 @@ repos:
     - id: check-yaml
     - id: end-of-file-fixer
     - id: trailing-whitespace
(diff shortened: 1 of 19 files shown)


francescofarina:
We're already using `ruff` which handles the sorting (even though I see it wasn't applied to some files). Can we configure `ruff` instead of introducing also `isort`? See [here](https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort)

TristanBilot:
Honestly I tried to use ruff's import sorting but I can't make it work I don't get why.  I saw than PyG is also using ruff+isort.

TristanBilot:
ok it's all good now

PR #73, opened 2024-02-08
Base Dataset class 

francescofarina:
## Proposed changes

Introducing the base `Dataset` class to build specific datasets upon. `KarateClubDataset` is implemented as a simple first dataset.

## Checklist


TristanBilot:
Amazing! Seems good to me!

francescofarina:
Let's merge this then and iterate on it as we add more datasets.

PR #74, opened 2024-02-08
Build docs on release

francescofarina:
## Proposed changes

Documentation is currently built and published every time we push to main.
Since we're now releasing the package is probably best to build and publish the new docs only on new releases


TristanBilot (approved):
I agree!

PR #81, opened 2024-02-09
More datasets (QM7b, TU)

francescofarina:
## Proposed changes

Implementing more datasets.

## Checklist

Put an `x` in the boxes that apply.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation

PR #72, opened 2024-02-07
Benchmarks (scatter/gather)

TristanBilot:

PR #68, opened 2024-02-02
WIP: a first notebook tutorial

TristanBilot:
## Proposed changes

Support for tutorial guides, embedded directly within notebooks.

PR #83, opened 2024-02-11
release 0.0.3

### LINKED ISSUE #88: add pandoc to docs build workflow

francescofarina:
## Proposed changes

Preparing for the new release.

This requires to first merge #68 #72 #81 and mlx new release (including the fast scatter op). 

Close #88 

## Checklist


francescofarina:
building docs is currently broken for some reason - I'll look into it once we merge #85 #87 into main as that may solve it.

francescofarina:
Building docs is now fixed - @TristanBilot whenever we need to import optional packages (that are not specified in the requirements) let's do it in the functions where that's necessary and not in the import statements at the beginning (see how I modified qm7b.py)

benchmarks/framework_comparison/training_loop/launcher.py
```diff
@@ -0,0 +1,142 @@
...
+
+FRAMEWORKS = [
+    # "dgl",
+    # "pyg",
+    "mxg",
+]
+DATASETS = [
+    # "BZR_MD",
+    # "MUTAG",
+    # "DD",
```
TristanBilot: shall we uncomment these? (frameworks + datasets)

benchmarks/framework_comparison/training_loop/launcher.py
```diff
@@ -0,0 +1,142 @@
...
+                        f"{layer_name}",
+                        f"{time:.3f}s",
+                    ]
+                )
+
+print(f"\nPlatform {platform.platform(terse=True)}")
+print(f"\nmlx version: {version('mlx')}")
+print(f"\nmlx-graphs version: {version('mlx_graphs')}")
+print(f"\ntorch version: {version('torch')}")
+print(f"\ntorch_geometric version: {version('torch_geometric')}")
```
TristanBilot: we should add the DGL version here too.

benchmarks/profiling/profile_training.py
```diff
@@ -0,0 +1,123 @@
...
+    )
+    optimizer.update(model, grads)
+    return loss
+
+
+def train(loader):
+    loss_sum = 0.0
+    for graph in loader:
+        loss = step(graph)
+        # mx.eval(model.parameters(), optimizer.state)
```
TristanBilot: remove this comment unless we use `mx.compile` in this script.

README.md
```diff
@@ -37,15 +33,73 @@ pip install mlx-graphs
...
-### Contributing
-#### Installing test, dev, benchmaks, docs dependencies
+## Usage
+
+### Tutorial guides
+
+We provide some notebooks to practice `mlx-graphs`.
+
+- Quickstart guide: [notebook](https://mlx-graphs.github.io/mlx-graphs/tutorials/quickstart.html)
+- Graph classification guide: [notebook](https://mlx-graphs.github.io/mlx-graphs/tutorials/graph_classification.html)
```
TristanBilot, line 50: I made the assumption that those URLs will work, can you confirm they are right @francescofarina?

TristanBilot:
Ah also, can you please run the `graph_classification.ipynb` notebook on your M3 Pro? The prints are not finished in the notebook and the results will be better than on my M1 Pro

TristanBilot:
Sorry I meant the `benhcmark_pyg_dgl_mxg.ipynb` notebook!

francescofarina:
Done! That notebook is not in the docs though (and the actual benchmark is in the benchmarks folder).

TristanBilot:
I think it would be worth it to add it to the docs no?

francescofarina:
> I think it would be worth it to add it to the docs no?

We can add it but it could be worth adding some brief commentary to it (and removing the compile step), wdyt? The way I see it useful in the docs is if we use it as a way to show the difference wrt pyg or dgl in terms of syntax/usage.

TristanBilot:
Actually for the `raph_classification.ipynb` it may be better to keep the last version as the accuracy was higher 😅

TristanBilot:
> > I think it would be worth it to add it to the docs no?
> 
> We can add it but it could be worth adding some brief commentary to it (and removing the compile step), wdyt? The way I see it useful in the docs is if we use it as a way to show the difference wrt pyg or dgl in terms of syntax/usage.

I thought this notebook was the "proof" that `mlx-graphs` can be faster than other frameworks, no? It is especially important IMO. Maybe in this release it would be enough to add it as is?

francescofarina:
I think notebooks in the docs should be _instructional_. 
Benchmarks should stay in the benchmarks folder in general IMO.

We can maybe add just this one to the docs for now. Let me add some commentary to it.

> Actually for the graph_classification.ipynb it may be better to keep the last version as the accuracy was higher 😅
Same comment as above, mostly instructional - I reduced the epochs from 100 to improve the readability.

francescofarina:
@TristanBilot I updated the nb and added it to the docs. I also added a note in the README and the first page of the docs saying that in initial benchmarks we're up to 10x faster than other frameworks - let me know what you think.

PR #87, opened 2024-02-13
improve get_unique_indices + refactoring

### LINKED ISSUE #86: Remove/add self loops extremely slow
This function is so slow that we can't use it in practice, maybe we can take inspiration from [PyG](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/utils/loop.html#remove_self_loops).

To reproduce:
```python
import mlx.core as mx
from mlx_graphs.utils import remove_self_loops

edge_index = mx.ones((2, 1_000_000))
remove_self_loops(edge_index)
```

I ran this 3min ago and it's still running 😅

I also get this once it's finished: ValueError: [gather] Got indices with invalid dtype. Indices must be integral.

francescofarina:
## Proposed changes

Closes #86.

It now takes 0.08s to remove self-loops from an edge index with all ones and 1,000,000 edges. I had to momentarily resort to numpy.

See also inline comment below.

I also made some refactoring to avoid circular imports.

## Checklist


mlx_graphs/utils/transformations.py
```diff
@@ -192,14 +192,17 @@ def get_unique_edge_indices(edge_index_1: mx.array, edge_index_2: mx.array) -> m
...
     """
-    edge_1_tuples = [tuple(edge.tolist()) for edge in edge_index_1.transpose()]
-    edge_2_set = set(tuple(edge.tolist()) for edge in edge_index_2.transpose())
+    edge_2_unique = np.unique(edge_index_2.transpose(), axis=0).tolist()
 
     return mx.array(
-        [i for i, edge in enumerate(edge_1_tuples) if edge not in edge_2_set]
+        [
+            i
+            for i, edge in enumerate(np.array(edge_index_1.transpose(), copy=False))
```
francescofarina, line 203: it's quite weird that here if I do 

```
for i, edge in enumerate(edge_index_1.transpose()):
    ....
```
it's exponentially slower

francescofarina: See https://github.com/ml-explore/mlx/issues/680

TristanBilot:
I just ran your code locally, and it's very fast. However, it returns a tuple `(mx.array, None)`:

```python
>>> edge_index = mx.random.normal((2, 1_000_000))
>>> remove_self_loops(edge_index)
(array([[-0.632783, 0.0148992, -0.796312, ..., 0.485181, -1.36464, -0.452573],
       [-1.52533, 1.23279, 1.06905, ..., -0.712059, 0.0971604, 0.0981964]], dtype=float32), None)
```

TristanBilot:
I think it should more intuitive if we return only the `edge_index` when `edge_features` are not provided, wdyt?

francescofarina:
@TristanBilot Yes, I refactored also other functions that were using the same pattern to do the same.

francescofarina:
Merging as it contains some refactoring useful for other PRs.

PR #89, opened 2024-02-15
Benchmark before release

TristanBilot:


TristanBilot:
There's something that `@cached` may broke:
```python
@cached
    def num_nodes(self) -> Union[int, None]:
        """Number of nodes in the graph."""
        if self.node_features is not None:
            return self.node_features.shape[0]

        # NOTE: This may be slow for large graphs
        elif self.edge_index is not None:
            return np.unique(self.edge_index).size
        return None
```
If either `node_features` or `edge_index` is modified, the cache becomes outdated, meaning that we should also update the cache when one of those is modified by their setter. wdyt @francescofarina? It should become quite heavy

francescofarina:
In which cases do you think we'll need to modify `node_features` or `edge_index`?

Also @TristanBilot have you considered using `functools.cached_property` instead of writing a custom cache function?

TristanBilot:
If we do that, all attributes should be immutable. This limits the usage of the `GraphData` class. We have to make this decision. And `functools.cached_property` does not allow to invalidate the cache using setters like in here.

francescofarina:
Yes, I'm thinking in which cases they'd need to be mutable - I'm not sure we should allow that.

francescofarina:
Ok, so since we saw that introducing caching doesn't provide any benefits I'll revert this to [4d5b174](https://github.com/mlx-graphs/mlx-graphs/pull/89/commits/4d5b174e69596f42fa082fa05bef5e7621e3a003)

francescofarina:
@TristanBilot for some reason on my laptop the compiled mlx training step is slower on larger datasets, see [here](https://github.com/mlx-graphs/mlx-graphs/blob/93c9503dc7c085da83595dfdceb4a2306fcf35b2/benchmarks/framework_comparison/training_loop/results/M3Pro_5P_6E_14GPU_18G.md)

TristanBilot:
Strange, maybe can you show that in https://github.com/mlx-graphs/mlx-graphs/issues/92 so that awni tries on M1 Max?

Issue #92, opened 2024-02-17
Benchmarks on M1 Max


awni:
Benchmarking MLX vs PT graphs on an M1 Max:

Findings:
1 . MLX is faster on every case compared to PT
2. Compiling helps a lot but only on the bigger dataset (it's 2x in one case on the bigger dataset 🚀 )
3. Suspect issues with smaller graphs are in the scatter still just based on some light profiling

### PT:
```
BZR_MD
==========
GCNConv | 0.190s
GATConv | 0.312s

MUTAG
==========
Downloading https://www.chrsmrrs.com/graphkerneldatasets/MUTAG.zip
Extracting .mlx_graphs_data/MUTAG/MUTAG/MUTAG.zip
Processing...
Done!
GCNConv | 0.046s
GATConv | 0.067s

DD
==========
Downloading https://www.chrsmrrs.com/graphkerneldatasets/DD.zip
Extracting .mlx_graphs_data/DD/DD/DD.zip
Processing...
Done!
GCNConv | 3.930s
GATConv | 5.088s
```

### MLX No Compile:

```
BZR_MD
==========
GCNConv | 0.088s
GATConv | 0.114s

MUTAG
==========
GCNConv | 0.027s
GATConv | 0.035s

DD
==========
GCNConv | 2.026s
GATConv | 3.027s
```

### MLX Compiled

```
BZR_MD
==========
GCNConv | 0.083s
GATConv | 0.111s

MUTAG
==========
GCNConv | 0.025s
GATConv | 0.033s

DD
==========
GCNConv | 1.267s
GATConv | 1.513s
```

TristanBilot:
Thanks for the benchmark @awni, this is very interesting!
The M1 Max is clearly a game changer here compared to the M1 Pro and M3 Pro we already benchmarked. How many GPU cores and RAM do you have?

The compiled functions on the large graph are very neat, it seems clearly worth it to keep compiled functions by default in the framework, what do you think @awni ?

I think those results are sufficient for a first official release 🚀

awni:
>  it seems clearly worth it to keep compiled functions by default in the framework, what do you think @awni ?

For MLX Graphs I do think you should probably default compile the graph training step update. The only reason not to is if the input sizes change shape from batch to batch but it does not look like that is the case for you.

[...]

PR #101, opened 2024-02-21
WIP: experiments on batch padding

TristanBilot:
This PR aims to experiment if padding + JIT compiling can improve speedup performance.
Come from https://github.com/mlx-graphs/mlx-graphs/issues/92#issuecomment-1956939328.


francescofarina:
@TristanBilot shall we make the padding the default? What do you think?

PR #102, opened 2024-02-22
WIP: a first neighbor sampling loader

TristanBilot:
We're still missing many functions to implement an efficient neighbor sampling, but here are some trials on the implementation.


francescofarina:
Closing as it's carried over in #112

Issue #95, opened 2024-02-20
Add GNNBenchmark dataset


TristanBilot:
This [dataset](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.datasets.GNNBenchmarkDataset.html#torch_geometric.datasets.GNNBenchmarkDataset) would be valuable to integrate within the library.

PR #104, opened 2024-02-24
Superpixel datasets

francescofarina:
## Proposed changes

Implementing the MNIST and CIFAR10 datasets in #95.

## Checklist


mlx_graphs/utils/array_ops.py
```diff
@@ -73,3 +73,22 @@ def one_hot(labels: mx.array, num_classes: Optional[int] = None) -> mx.array:
...
+
+    Returns:
+        Array of shape (N, M)
+    """
+    assert x.shape[1] == y.shape[1], "Input vectors must have the same dimensionality"
+    # Use broadcasting to compute pairwise differences
+    expanded_x = mx.expand_dims(x, 1)
+    distances = mx.linalg.norm(expanded_x - y, axis=-1)
+
+    return distances
```
francescofarina, line 94: This is not too slow, similar to `cdist` in `scipy`.

mlx_graphs/datasets/superpixel.py
```diff
@@ -0,0 +1,214 @@
...
+
+def sigma(distances: mx.array, k: int = 8) -> mx.array:
+    """
+    Computes the scale parameter sigma defined as the averaged distance xk of the k
+    nearest neighbors for each node in the distances matrix.
+
+    See Equation (47) in `<http://arxiv.org/abs/2003.00982>`_ for more details.
+
+    Args:
+        distances:
```
TristanBilot, line 30: missing docs here

mlx_graphs/datasets/superpixel.py
```diff
@@ -0,0 +1,214 @@
...
+        adjacency_matrix = mx.exp(-((coord_dist / sigma(coord_dist)) ** 2))
+
+    # convert to symmetric matrix without self-loops
+    adjacency_matrix = 0.5 * (adjacency_matrix + adjacency_matrix.T)
+    for i in range(adjacency_matrix.shape[0]):
+        adjacency_matrix[i, i] = 0
+    return adjacency_matrix
+
+
+def compute_edges_list(
```
TristanBilot, line 84: maybe the name of this function could be more self-explanatory

mlx_graphs/datasets/superpixel.py
```diff
@@ -0,0 +1,214 @@
...
+
+    if num_nodes > k:
+        knns = mx.argpartition(adjacency_matrix, new_kth - 1, axis=-1)[:, new_kth:-1]
+        knn_values = mx.partition(adjacency_matrix, new_kth - 1, axis=-1)[:, new_kth:-1]
+    else:
+        # for graphs with less than k nodes the resulting graph will be fully connected.
+        knns = mx.repeat(mx.arange(num_nodes), num_nodes).reshape(num_nodes, num_nodes)
+        knn_values = adjacency_matrix
+
+        # TODO: eventually remove self-loops
```
TristanBilot, line 108: do you need to remove self-loops here?

mlx_graphs/datasets/utils/download.py
```diff
@@ -136,14 +142,51 @@ def check_sha1(filename: str, sha1_hash: str) -> bool:
     return sha1.hexdigest() == sha1_hash
 
 
-def extract_zip(path: str, folder: str, log: bool = True):
-    r"""Extracts a zip archive to a specific folder.
+def extract_archive(file, target_dir, overwrite=True):
```
TristanBilot, line 145: type hints missing

TristanBilot (changes requested):

TristanBilot:
Shall we add tests for this dataset too as in [here](https://github.com/mlx-graphs/mlx-graphs/blob/main/tests/datasets/test_qm7b.py) ?

mlx_graphs/datasets/superpixel.py
francescofarina: in practice it seems the first part of the if is the that's always executed for this two datasets - so I avoided adding the self-loop removal as it requires numpy for the moment (because we can't index with booleans in mlx). I'll add that part anyways though.

francescofarina:
> Shall we add tests for this dataset too as in [here](https://github.com/mlx-graphs/mlx-graphs/blob/main/tests/datasets/test_qm7b.py) ?

Added - it tests shape of the `edge_index` for each batch and the `graph_labels`.

tests/datasets/test_superpixel.py
```diff
@@ -0,0 +1,37 @@
...
+
+import mlx.core as mx
+import pytest
+
+from mlx_graphs.datasets import SuperPixelDataset
+from mlx_graphs.loaders import Dataloader
+
+
+@pytest.mark.slow
+def test_tu_dataset():
```
TristanBilot, line 12:
```suggestion
def test_superpixel_dataset():
```

TristanBilot:
> Added - it tests shape of the `edge_index` for each batch and the `graph_labels`.

Perfect! Did you verify that the yielded images are identical compared to the original dataset? Why comparing the shapes instead of the full `edge_index`?

francescofarina:
There are some differences in how the edges are ordered - but let me find a way to make a tight comparison just to be sure

francescofarina:
So quick update: it seems there are some numerical errors between how we compute pairwise distances and `cdist` from `scipy` which cause our dataset to be slightly different. By running the test in https://github.com/mlx-graphs/mlx-graphs/blob/0ae136aca7bc4d02cb9792fe8f6ae56540853500/tests/utils/test_array_ops.py we can see that there are some numerical differences of about 1e-7 in some points. I can either go with `cdist` (which however will require depending on `scipy`) or explore this further.

TristanBilot:
I'm not sure it's big deal if there's this small 1e-7 offset. Do you think this can have an influence on the overall training?

francescofarina:
Yes, unfortunately it does - those small differences influences the knn neighbors that are used to compute the superpixels (so the `edge_index` we have for each graph is slightly different from the one of PyG or dgl).

TristanBilot:
Oh ok I see. Do you know where the numerical error comes from? As you basically only did:

```python
distances = mx.linalg.norm(expanded_x - y, axis=-1)
```
Maybe we can go with `cdist` for now, and open an issue on this.

francescofarina:
So, turns out that the major problem is the output of `argpartition` in `adjacency_matrix_to_knn_edges`. What happens is that when there are equal elements in a partition, their order is random with mlx, while when using `numpy` they're ordered based on their index in the original array. This turns out to be a significant difference with highly connected graphs.

If you're curious you can see it in action by running

```python
import numpy as np
import mlx.core as mx

mx.random.seed(1)

A = mx.random.randint(0,10,[10,10])
A_np = np.array(A)

print(A_np)
>>> [[2 9 2 9 9 9 1 7 9 7]
 [1 0 5 4 5 0 8 4 5 4]
 [0 7 8 7 8 6 4 6 2 8]
 [4 5 6 1 4 5 7 1 7 1]
 [4 5 2 9 6 7 6 1 0 9]
 [0 3 5 2 6 3 5 4 2 2]
 [4 7 4 9 4 6 4 6 8 0]
 [9 0 7 5 9 1 1 8 8 7]
 [4 1 7 8 9 8 9 2 6 9]
 [1 8 3 5 0 0 1 1 1 4]]

print(np.array(mx.argpartition(A, 5, axis=-1).tolist()))
>>> [[6 0 2 7 9 1 3 4 5 8]
 [1 5 0 3 7 9 2 4 8 6]
 [0 8 6 5 7 1 3 2 4 9]
 [3 7 9 0 4 1 5 2 6 8]
 [8 7 2 0 1 4 6 5 3 9]
 [0 3 8 9 1 5 7 2 6 4]
 [9 0 2 4 6 5 7 1 8 3]
 [1 5 6 3 2 9 7 8 0 4]
 [1 7 0 8 2 3 5 4 6 9]
 [4 5 0 6 7 8 2 9 3 1]]

print(np.argpartition(A_np, 4, axis=-1))
>>> [[6 0 2 7 9 5 1 3 8 4]
 [1 0 7 5 9 3 6 2 8 4]
 [5 6 8 0 7 1 3 4 2 9]
 [3 9 7 0 4 1 5 2 8 6]
 [8 7 2 0 1 4 6 5 3 9]
 [8 0 9 3 1 5 7 6 2 4]
 [6 9 0 2 4 5 1 7 8 3]
 [1 5 6 3 9 2 8 7 0 4]
 [0 7 1 8 2 5 3 4 6 9]
 [4 5 8 7 6 0 1 3 2 9]]
```

By using `numpy` (arg)partition instead of `mlx` the number of different edges is 1,036 out of 5,645,128 in the MNIST test dataset (down from 538,170 with mlx argpartition).

Still figuring out why 1,036 edges are still different.

francescofarina:
I think there are multiple valid representations for this dataset due to how the superpixels are created.

The PyG and DGL versions are also different. @TristanBilot I think we can move forward with what we have.

PR #105, opened 2024-02-25
GCNConv self loops

nkanak:
## Proposed changes

Enable self-loops for the GCNConv layer.

## Checklist

### DIFF
--- a/mlx_graphs/nn/conv/gcn_conv.py
+++ b/mlx_graphs/nn/conv/gcn_conv.py
@@ -4,7 +4,7 @@ import mlx.core as mx
 import mlx.nn as nn
 
 from mlx_graphs.nn.message_passing import MessagePassing
-from mlx_graphs.utils import degree, invert_sqrt_degree
+from mlx_graphs.utils import add_self_loops, degree, invert_sqrt_degree
 
 
 class GCNConv(MessagePassing):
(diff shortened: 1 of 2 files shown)


mlx_graphs/nn/conv/gcn_conv.py
```diff
@@ -14,19 +14,22 @@ class GCNConv(MessagePassing):
...
         out_features_dim: int,
         bias: bool = True,
+        add_self_loops: bool = False,
         **kwargs,
     ):
         kwargs.setdefault("aggr", "add")
         super(GCNConv, self).__init__(**kwargs)
 
         self.linear = nn.Linear(node_features_dim, out_features_dim, bias)
+        self.add_self_loops = add_self_loops
```
francescofarina, line 32: Shall we make this private?

francescofarina (changes requested):

mlx_graphs/nn/conv/gcn_conv.py
francescofarina:
```suggestion
        self._add_self_loops = add_self_loops
```

nkanak: Ready!

francescofarina (approved):
Looks good 🚢

PR #108, opened 2024-02-27
MLX 0.4.0 benchmarks

TristanBilot:
Added M1 Pro results with the faster scatter introduced in MLX 0.4.0 + move benchs into separate folders with the mlx version. We need to find a way to also integrate mlx-graphs version (maybe as a prefix in the file name).


francescofarina:
@TristanBilot you may want to upgrade torch and pyg as well.

TristanBilot:
Nice results on the M3 Pro

Issue #94, opened 2024-02-20
Add neighbor sampling algorithm


TristanBilot:
Neighbor sampling as used in GraphSAGE is not implemented yet.

francescofarina:
@Kheims will you be working on this?

PR #109, opened 2024-02-27
Convert to nx

Kheims:
## Proposed changes

Simple utility function to convert an mlx GraphData object to a networkx object to facilitate visualisation of subgraphs of Neighbor Sampler (#94)

## Checklist

### DIFF
--- a/docs/source/api/utils/convert.rst
+++ b/docs/source/api/utils/convert.rst
@@ -1,0 +1,14 @@
+.. _convert:
+
+
+Convert
+=========
+
+
+.. currentmodule:: mlx_graphs.utils.convert
+
+
+.. autosummary::
+	:toctree: _autosummary_convert
+
+	to_networkx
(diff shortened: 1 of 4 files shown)


Kheims:
Note : I haven't added conversion of graphs intrinsic attributes but only relevant structural information for strictly visualisation purposes.

TristanBilot:
Nice initiative @Kheims 🚀 I just pushed some minor fixes. Btw, you don't need to add the type hints in the function docstring as it's automatically infered by Sphinx based on the function signature.

TristanBilot:
Also, @francescofarina I think it would be better to return 0 is `num_nodes` instead of None when there's no nodes, wdyt?

Kheims:
Oh i didn't know that, thanks for the info and fixes

francescofarina:
> Also, @francescofarina I think it would be better to return 0 is `num_nodes` instead of None when there's no nodes, wdyt?

Yes, otherwise we have to always make the check to avoid type hinting issues.

PR #112, opened 2024-02-28
WIP: neighbor loader first implem

Kheims:
## Proposed changes

Here is a first implementation of Neighbor Sampling.

The idea is :
we have a collate_fn, namely `sample_graph ` that from an original graph randomly samples iteratively neighbors of a seed node, and the neighbors of neighbors according to the `num_neighbors` list, returning along the built sampled_edge, some references necessary for later use in Neighbor Loader, namely, `n_id` : the references to the sampled_nodes, `e_id` : references to edges built from sampled nodes and `input_nodes` : reference to seed nodes on which we will later compute the embedding.
Effectively, a `batch_size` might be specified, in this case we will sample nodes for a `batch_size` number of seed nodes, thus helping to build a single computational graph taking into account multiple seed nodes, if input_nodes are not specified `batch_size` number of seed nodes will be randomly selected.

NeighborLoader then use this collate_fn to iterate on the graph yielding subgraphs with associated references to original graph.

As you might see, alot of optimisation needs to be done, as there are some implicit conversions between numpy arrays and mx.arrays, and some for loops that can be alleviated.

Thus, this draft PR to discuss a little bit about the intricacies here and there to have a go as i engage to profile this implementation and optimize it as much as possible, until we shift to a cpp implementation.

PS : I have left a notebook where you can visualize original graph and it's sampled subgraph.

## Checklist

### DIFF
--- a/docs/source/_templates/transform-class-template.rst
+++ b/docs/source/_templates/transform-class-template.rst
@@ -1,0 +1,20 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+	:members:
+	:show-inheritance:
+
+	{% block methods %}
+	{% if methods %}
+	.. rubric:: {{ _('Methods') }}
+
+	.. autosummary::
+	{% for item in methods %}
+	  {%- if item not in inherited_members and item != '__init__' %}
+	     ~{{ name }}.{{ item }}
+	  {%- endif %}
+	{%- endfor %}
+	{% endif %}
+	{% endblock %}
--- a/docs/source/api/datasets/index.rst
+++ b/docs/source/api/datasets/index.rst
@@ -23,3 +23,4 @@ Datasets
 	TUDataset
 	SuperPixelDataset
 	OGBDataset
+	EllipticBitcoinDataset
--- a/docs/source/api/nn/index.rst
+++ b/docs/source/api/nn/index.rst
@@ -19,6 +19,9 @@ Neural Networks
 	GINConv
 	SAGEConv
 	GeneralizedRelationalConv
+	BatchNormalization
+	InstanceNormalization
+	LayerNormalization
 	global_add_pool
 	global_max_pool
 	global_mean_pool
--- a/docs/source/api/transforms/index.rst
+++ b/docs/source/api/transforms/index.rst
@@ -1,0 +1,14 @@
+.. _transforms:
+
+Transforms
+===============
+
+.. currentmodule:: mlx_graphs.transforms
+
+
+.. autosummary::
+	:toctree: _autosummary
+	:template: transform-class-template.rst
+
+	BaseTransform
+	FeaturesNormalizedTransform
--- a/docs/source/api/utils/convert.rst
+++ b/docs/source/api/utils/convert.rst
@@ -12,3 +12,4 @@ Convert
 	:toctree: _autosummary_convert
 
 	to_networkx
+	from_networkx
--- a/docs/source/api/utils/transformations.rst
+++ b/docs/source/api/utils/transformations.rst
@@ -19,3 +19,6 @@ Transformations
 	to_undirected
 	remove_duplicate_directed_edges
 	coalesce
+	get_isolated_nodes_mask
+	has_isolated_nodes
+	has_self_loops
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -2,14 +2,14 @@ import os
 import sys
 
 sys.path.insert(0, os.path.abspath("../.."))
-autodoc_mock_imports = ["mlx"]
+autodoc_mock_imports = ["mlx", "networkx"]
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 project = "mlx-graphs"
 copyright = "2024, mlx-graphs contributors"
 author = "mlx-graphs contributors"
-release = "0.0.4"
+release = "0.0.6"
 
 # -- General configuration ---------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
@@ -22,6 +22,7 @@ extensions = [
     "sphinx.ext.intersphinx",
     "nbsphinx",
     "sphinx_gallery.load_style",
+    "sphinx.ext.viewcode",
 ]
 
 napoleon_use_param = True
@@ -59,6 +60,7 @@ html_theme_options = {
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
     "mlx": ("https://ml-explore.github.io/mlx/build/html", None),
+    "networkx": ("https://networkx.org/documentation/stable", None),
 }
 
 html_css_files = [
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -90,6 +90,7 @@ Here is an example of a custom `GraphSAGE <https://proceedings.neurips.cc/paper_
    api/data/index.rst
    api/datasets/index.rst
    api/loaders/index.rst
+   api/transforms/index.rst
    api/nn/index.rst
    api/utils/index.rst
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -19,7 +19,7 @@ To build and install `mlx-graphs` from source start by cloning the github repo
 
 .. code-block:: shell
 
-	git clone git@github.com:TristanBilot/mlx-graphs.git && cd mlx-graphs
+	git clone git@github.com:mlx-graphs/mlx-graphs.git && cd mlx-graphs
 
 Then build and install using `pip` (we suggest using a separate virtual environment)
 
--- a/mlx_graphs/data/data.py
+++ b/mlx_graphs/data/data.py
@@ -3,6 +3,9 @@ from typing import Literal, Optional, Union
 import mlx.core as mx
 import numpy as np
 
+from mlx_graphs.utils import has_isolated_nodes, has_self_loops
+from mlx_graphs.utils.topology import is_undirected
+
 
 class GraphData:
     """
@@ -28,7 +31,7 @@ class GraphData:
 
     def __init__(
         self,
-        edge_index: Optional[mx.array] = None,
+        edge_index: mx.array,
         node_features: Optional[mx.array] = None,
         edge_features: Optional[mx.array] = None,
         graph_features: Optional[mx.array] = None,
@@ -70,23 +73,18 @@ class GraphData:
         return {k: v for k, v in self.__dict__.items() if v is not None}
 
     @property
-    def num_nodes(self) -> Union[int, None]:
+    def num_nodes(self) -> int:
         """Number of nodes in the graph."""
         if self.node_features is not None:
             return self.node_features.shape[0]
-
-        # NOTE: This may be slow for large graphs
-        elif self.edge_index is not None:
-            return np.unique(self.edge_index).size
-        return None
+        else:
+            # NOTE: This may be slow for large graphs
+            return np.unique(np.array(self.edge_index, copy=False)).size
 
     @property
-    def num_edges(self) -> Union[int, None]:
+    def num_edges(self) -> int:
         """Number of edges in the graph"""
-        if self.edge_index is not None:
-            return self.edge_index.shape[1]
-
-        return None
+        return self.edge_index.shape[1]
 
     @property
     def num_node_classes(self) -> int:
@@ -173,3 +171,20 @@ class GraphData:
         if "index" in key:
             return self.num_nodes
         return None
+
+    def has_isolated_nodes(self) -> bool:
+        """Returns a boolean of whether the graph has isolated nodes or not
+        (i.e., nodes that don't have a link to any other nodes)"""
+        return has_isolated_nodes(self.edge_index, self.num_nodes)
+
+    def has_self_loops(self) -> bool:
+        """Returns a boolean of whether the graph contains self loops."""
+        return has_self_loops(self.edge_index)
+
+    def is_undirected(self) -> bool:
+        """Returns a boolean of whether the graph is undirected or not."""
+        return is_undirected(self.edge_index, self.edge_features)
+
+    def is_directed(self) -> bool:
+        """Returns a boolean of whether the graph is directed or not."""
+        return not self.is_undirected()
--- a/mlx_graphs/datasets/__init__.py
+++ b/mlx_graphs/datasets/__init__.py
@@ -5,3 +5,4 @@ from .tu_dataset import TUDataset  # noqa
 from .planetoid import PlanetoidDataset  # noqa
 from .superpixel import SuperPixelDataset  # noqa
 from .ogb_dataset import OGBDataset  # noqa
+from .elliptic import EllipticBitcoinDataset  # noqa
--- a/mlx_graphs/datasets/dataset.py
+++ b/mlx_graphs/datasets/dataset.py
@@ -2,7 +2,7 @@ import copy
 import os
 import pickle
 from abc import ABC, abstractmethod
-from typing import Literal, Optional, Sequence, Union
+from typing import Callable, Literal, Optional, Sequence, Union
 
 import mlx.core as mx
 import numpy as np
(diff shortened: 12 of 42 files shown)


Kheims:
I'm gonna add tests for the collate_fn and it's edge cases, but how do you see tests for the neighbor loader ?

francescofarina:
@Kheims what's the difference wrt #102?

francescofarina:
> I'm gonna add tests for the collate_fn and it's edge cases, but how do you see tests for the neighbor loader ?

To start we can test whether batching and shapes of the sampled graphs are as expected, wdyt?

Kheims:
> > I'm gonna add tests for the collate_fn and it's edge cases, but how do you see tests for the neighbor loader ?
> 
> To start we can test whether batching and shapes of the sampled graphs are as expected, wdyt?

For the shapes i agree, but for the batching we can't know in advance the output subgraphs as it is not a deterministic process ? or should we use a seed

Kheims:
> @Kheims what's the difference wrt #102?

I have rewritten the sample_nodes from scratch, but indeed no difference wrt to the addressed issue

mlx_graphs/loaders/neighbor_loader.py
```diff
@@ -0,0 +1,44 @@
+from typing import List
+
+import mlx.core as mx
+import numpy as np
+
+from mlx_graphs.data import GraphData
+from mlx_graphs.sampler.neighbor_sampler import sample_nodes
+
+
+class NeighborLoader:
```
francescofarina, line 10: I'm not sure whether this should be a data loader or just a sampler that we can use to generate samples to be loaded through the current `Dataloader`. @TristanBilot @Kheims what do you think?

Kheims: Hmm, imo you're right we should directly use the DataLoader, but we should therefore provide to the data loader a sampler function which means some modifications needs to done in it ?

francescofarina: It can also be a two-stage process, like:
```python
graph = GraphData(...)
samples = sampler(graph, num_samples)
dl = Dataloader(samples, batch_size)
```

Kheims: Seems good to me

TristanBilot: Interesting. Yeah I find that using a dedicated `DataLoader` everytime we want to sample is suboptimal. We should provide both a `sample` function and a `NeighborSampler` class which can be added as argument to every `DataLoader`. 
However, we have to check also the compatibility between the graph batching of the `DataLoader` and the node batching of the `NeighborSampler`.

francescofarina: It's the same approach of [GraphSAGE](https://arxiv.org/pdf/1706.02216.pdf), right?

Kheims: Yes absolutely, the question is just to make sure about the implementation details (not cited in the original paper but done in implementations of pyg and dgl ( but they rely on an cpp implementation))

francescofarina: Yes, that seems correct to me - you can create a dummy test with a graph like
```python
#   0  1
#  / \/ \
#  2  3  4
#  |  |  |
#  5  6  7
```
where the sampling should be deterministic (see the [example in pyg](https://pytorch-geometric.readthedocs.io/en/latest/tutorial/neighbor_loader.html)).

Kheims: In the example they use a batch size of 1 where we expect to have dissociated subgraphs for each input nodes, but if batch size is 2 with input nodes 0 and 1 then we should associate the subgraphs to contain a single computational graph containing both subgraphs ( in this case all the graph you mentioned ) 
This is a straightforward example but i wanted a more complex one to test robustly the functionality ( where we had choices in sampling in each hop )

francescofarina: @Kheims I'm not sure I understand - when batch size is grater than 1, shouldn't you just repeat the sampling procedure and then collate the resulting subgraphs?

francescofarina: @Kheims I think you're right - my understanding is that the procedure you were suggesting is the correct one, rather than just sampling and collating the results. @TristanBilot do you agree?

TristanBilot:
I think that performing sampling on a batch of seed nodes should be as easy as repeating the sampling function on each node and collate them into a single graph. 

This is nicely explained in [Jure Leskovec's video](https://www.youtube.com/watch?v=LLUxwHc7O4A):
image

TristanBilot:
btw it should be quite easy to implement the multi-processing version of the sampler by spawning a separate process per seed node. We have to check whether it's better to run the operations on CPU or GPU in this scenario though.

francescofarina:
@Kheims any updates on this?

Kheims:
> @Kheims any updates on this?

Yes, i was just thinking about a clever way to collate the subgraphs, in the sampler you can see that we iterate over all seed nodes and sample through the neighbor sampler all required attributes of the subgraphs, if a batch size of 2 is given for exemple, in the subgraphs list we would iterate over this list 2 elements by 2 and collate them into a single graph ( straightforward, just append the other edges and remove doubles as embedding is already computed from a single edge)
wdyt ? 
+ Should we order edge_index and follow a certain format ?

TristanBilot:
In a first time, collating graphs should be quite easy as it's O(|E|), you just have to merge duplicate edges. Once a greedy approach is implemented, you can think about how we can vectorize or parallelize this solution. (Btw, do you really have to collate? If the 2 graphs you wanna batch are defined with their node IDs then a concatenation like in batching is sufficient to collate).

Kheims:
> In a first time, collating graphs should be quite easy as it's O(|E|), you just have to merge duplicate edges. Once a greedy approach is implemented, you can think about how we can vectorize or parallelize this solution. (Btw, do you really have to collate? If the 2 graphs you wanna batch are defined with their node IDs then a concatenation like in batching is sufficient to collate). 
> 

It depends if we want to enforce some ordering of the edge_index therefore reordering for the required references should be made from my perspective.
But indeed a concatenation is sufficient in normal cases, the usage of the term collate was out of context ^^'

TristanBilot:
For the moment, there's no ordering for the edge_index. In the future, we will certainly sort the edge_index in some cases to accelerate scattering operations but currently there's no ordering.

Kheims:
I was wondering if providing a utility function to randomly create a graph with specific structure and number of nodes and connection would be relevant ? In my case it's to profile my sampler.
Is there a plus value or should we(and user) just stick to initialize the attributes as we want ? @TristanBilot @francescofarina

TristanBilot:
I'm not sure there's a huge value in integrating such a function as it's very straightforward to implement. Maybe for test utils?
[Here](https://github.com/TristanBilot/mlx-benchmark/blob/main/mlx_benchmark/utils.py#L98) is an example for generating a dummy `edge_index`.

Kheims:
> I'm not sure there's a huge value in integrating such a function as it's very straightforward to implement. Maybe for test utils?
> 
> [Here](https://github.com/TristanBilot/mlx-benchmark/blob/main/mlx_benchmark/utils.py#L98) is an example for generating a dummy `edge_index`.

But it doesn't enforce specific connections for the edge_index we just might get a dag out of it in the worst case ?

francescofarina:
Perhaps we can generate and convert it from `networkx` for now?

TristanBilot (changes requested):
Got these errors when running your tests @Kheims:
```
FAILED sampler/test_neighbor_sampler.py::test_sample_nodes_no_hop - TypeError: sample_nodes() got an unexpected keyword argument 'batch_size'
FAILED sampler/test_neighbor_sampler.py::test_sample_nodes_one_hop - TypeError: sample_nodes() got an unexpected keyword argument 'batch_size'
FAILED sampler/test_neighbor_sampler.py::test_sample_nodes_two_hop - TypeError: sample_nodes() got an unexpected keyword argument 'batch_size'
```
- Also, can you rename the `sampler/` and `tests/sampler/` by `sampling/` to be consistent with naming.
- It would be nice to have a proper profiling of the sampler, can you run multiple experiments with large shapes and measure runtime? 
- Can you add additional tests with edge cases (e.g. `num_neighbors=[0]`, `num_neighbors=[-1]`, etc)

Kheims:
I've added profiling, sampler execution time for the Cora dataset and **100 input nodes** from which to sample **[10, 2]** neighbors takes `0.05 seconds`, it is interesting to take a look at the line_profiling output specifically 
`mask = (structured_edge_index.source == node) & (~np.isin(structured_edge_index.target, list(visited_nodes))` takes almost 60% of the execution time of the `sample_nodes` function which is naturally the bottleneck in this case where it takes 80% of the `sampler` function

I need to further inspect the outputs, as for now I am using random indices for the input nodes from the edge_index, so it might take nodes with low degree, I have to provide nodes with highest degrees to have a better grasp on the worst cases scenarios.

TristanBilot:
Interesting benchmarks, thanks!
In a real-world scenario, we'll likely using all the available nodes in the dataset as seed nodes though, so I updated your bench accordingly. Also, I found an error in the creation of your `input_nodes`. You generated nodes in the range `1...num_edges+1` instead of `0...num_nodes`. This reduced the runtime a lot in most cases where the graph has more edges than nodes, as there will be no sampling to perform when your randomly-generated node  IDs are greater than `num_nodes`.

Overall, it's taking ~35s to compute the sampling on all nodes of the `pubmed` dataset which comprises 19,717 nodes and 88,648 edges in avg. On the very small datasets `cora` and `citeseer`, it's taking ~1.5s.

To have an idea whether the current sampling is feasible, it would be nice to compare with PyG's sampler, could you try that @Kheims?

TristanBilot:
I just improved runtime by 6x on `pubmed` and ~2x on `cora` and `citeseer` by finding an alternative to the expensive `np.isin`.
New times: 7s on `pubmed`, 0.8s on `cora`, 0.74s on `citeseer`

TristanBilot:
btw:
- @Kheims the `NeighborLoader` is broken, could you fix it and write some tests when you have time after the PyG's benchmark?
- also, the default behavior of setting `num_neighbors=[-1]` should be to keep all the neighbors in the sampling, instead of returning an empty array in the tests.

Kheims:
> btw:
> 
> - @Kheims the `NeighborLoader` is broken, could you fix it and write some tests when you have time after the PyG's benchmark?
> 
> - also, the default behavior of setting `num_neighbors=[-1]` should be to keep all the neighbors in the sampling, instead of returning an empty array in the tests.
> 
> 

Ok for second.
For the neighbor loader we have agreed to rely on DataLoader to which we provide samples sampled through the sampler so Neighbor Loader is no longer of use

TristanBilot:
Yes, indeed! In this case we have to integrate the sampling in the DataLoader

TristanBilot:
Any updates? @Kheims Can you push your benchmark code and results?

Kheims:
To instantiate pyg's NeighborLoader with similar args as our implem benchmark, it takes :  0.0036s on `CiteSeer`, 0.0033 on `Cora` and 0.007s on `Pubmed`

francescofarina:
> To instantiate pyg's NeighborLoader with similar args as our implem benchmark, it takes : 0.0036s on `CiteSeer`, 0.0033 on `Cora` and 0.007s on `Pubmed`

I think we have to understand why there's a 1000x difference here. Also, I don't think we can use a sampler that takes 7s to generate a sample if we generate them on the fly.

Kheims:
> > To instantiate pyg's NeighborLoader with similar args as our implem benchmark, it takes : 0.0036s on `CiteSeer`, 0.0033 on `Cora` and 0.007s on `Pubmed`
> 
> 
> 
> I think we have to understand why there's a 1000x difference here. Also, I don't think we can use a sampler that takes 7s to generate a sample if we generate them on the fly.

Main reason is they have a cpp implementation of their sampler.
For the second point, from what we agreed upon, samples won't be generated on the fly but rather first generated then given to the vanilla dataloader ?

TristanBilot:
The bottleneck surely doesn't only come from the cpp implementation, we'll have to investigate the PyG's code in a future PR. In the meantime, I think this first version could be merged so that people can start play with it, and we'll work at the same time on a v2 which has to be implemented in cpp directly in the MLX library.

Kheims:
> The bottleneck surely doesn't only come from the cpp implementation, we'll have to investigate the PyG's code in a future PR. In the meantime, I think this first version could be merged so that people can start play with it, and we'll work at the same time on a v2 which has to be implemented in cpp directly in the MLX library.

It will be good also to start implementing GraphSage with this neighborsampler to see the limits and perhaps understand better the actual bottlenecks, i agree that it is not only due to the cpp implementation but i feel it's also due to how they manage all their graph's data structs using global storages

TristanBilot:
I cleaned some stuff and improved the API to be more intuitive. I think we could merge for this v1.
I figured out that some edges can be duplicated in the returned subgraph. Were you aware?

Example:

```python
edge_index = mx.array([[0, 0, 1, 1, 2, 3, 4], [2, 3, 3, 4, 5, 6, 7]])
node_features = mx.array([[1], [1], [1], [1], [1], [1], [1], [1]])
graph = GraphData(edge_index=edge_index, node_features=node_features)
subgraphs = sample_neighbors(
    graph=graph, num_neighbors=[-1, -1], input_nodes=[0, 1]
)
assert len(subgraphs) == 1
assert mx.array_equal(mx.sort(subgraphs[0].edge_index), mx.sort(edge_index))   # the edge (3, 6) is duplicated in subgraphs[0].edge_index
```

Kheims:
> I cleaned some stuff and improved the API to be more intuitive. I think we could merge for this v1. I figured out that some edges can be duplicated in the returned subgraph. Were you aware @Kheims ?
> 
> Example:
> 
> ```python
> edge_index = mx.array([[0, 0, 1, 1, 2, 3, 4], [2, 3, 3, 4, 5, 6, 7]])
> node_features = mx.array([[1], [1], [1], [1], [1], [1], [1], [1]])
> graph = GraphData(edge_index=edge_index, node_features=node_features)
> subgraphs = sample_neighbors(
>     graph=graph, num_neighbors=[-1, -1], input_nodes=[0, 1]
> )
> assert len(subgraphs) == 1
> assert mx.array_equal(mx.sort(subgraphs[0].edge_index), mx.sort(edge_index))   # the edge (3, 6) is duplicated in subgraphs[0].edge_index
> ```

Yes, I think I told you about it last time we went at Isep, I don't know the awaited dealing with this ?

TristanBilot:
Another fix that would be appreciated. This returns 3 graphs in the list instead of 2. The 2 first graphs are correct, but there is an additional empty graph at the end. This should be an easy fix.

```python
edge_index = mx.array([[0, 0, 1, 1, 2, 3, 4], [2, 3, 3, 4, 5, 6, 7]])
graph = GraphData(edge_index=edge_index)

subgraphs = sample_neighbors(
    graph=graph, num_neighbors=[1], batch_size=3,
)
```

francescofarina:
> I have difficulties to load the docs, is it usually enough to remove `build/` to make the docs building work @francescofarina? Removing all autosummary files manually is long lol.

What errors are you getting? If there are conflicts you may actually need to remove the autosummary folders that are causing them - that's why we currently have a different autosummary folder for each submodule.

mlx_graphs/sampling/neighbor_sampler.py
```diff
@@ -48,6 +48,23 @@ def sample_neighbors(
         edge_index = mx.array([[0, 0, 1, 1, 2, 3, 4], [2, 3, 3, 4, 5, 6, 7]])
         graph = GraphData(edge_index=edge_index)
 
+        #  Sample nodes 0 and 1 with two neighbors
```
TristanBilot, line 51: why did you revert my changes? 😅

mlx_graphs/sampling/neighbor_sampler.py
```diff
@@ -159,7 +159,7 @@ def sample_neighbors(
                 input_nodes=mx.array(input_node),
             )
             if graph.node_features is not None:
-                graph.node_features = graph.node_features[mx.array(batch_n_id)]
+                subgraph.node_features = graph.node_features[mx.array(batch_n_id)]
```
TristanBilot, line 162: good catch!

francescofarina:
@Kheims @TristanBilot just a check in on this - are you happy with the implementation?

TristanBilot:
> @Kheims @TristanBilot just a check in on this - are you happy with the implementation?

hmm happy, not really lol, but we can merge and improve the speedup later!

TristanBilot:
Closing this PR as already implemented by @vinayhpandya

PR #118, opened 2024-03-01
TUDataset fix

francescofarina:
## Proposed changes

Fixing the `TUDataset` as it wasn't loading node and edge attributes.

## Checklist


francescofarina:
I was playing with QM9 and noticed that attributes weren't loaded because of the wrong file names.

PR #130, opened 2024-03-04
ogb dataset

### LINKED ISSUE #127: Importer/converter of OGB datasets
See https://ogb.stanford.edu/docs/dataset_overview/. We should be able to use the library-agnostic datasets and convert them to the mlx-graphs format.

francescofarina:
## Proposed changes

Implementing OGB datasets. Closes #127 

## Checklist

### DIFF
--- a/mlx_graphs/datasets/ogb_dataset.py
+++ b/mlx_graphs/datasets/ogb_dataset.py
@@ -1,0 +1,115 @@
+from typing import Literal, Optional, Union, get_args, overload
+
+import mlx.core as mx
+import numpy as np
+
+from mlx_graphs.data.data import GraphData
+from mlx_graphs.datasets import Dataset
+from mlx_graphs.utils import index_to_mask
+
+OGB_NODE_DATASET = Literal[
+    "ogbn-products",
+    "ogbn-proteins",
+    "ogbn-arxiv",
+    "ogbn-papers100M",
+    # "ogbn-mag", # TODO: requires heterogeneous graphs
+]
+OGB_EDGE_DATASET = Literal[
+    "ogbl-ppa",
+    "ogbl-collab",
+    "ogbl-ddi",
+    "ogbl-citation2",
+    # "ogbl-wikikg2",  # TODO: requires heterogeneous graphs
+    # "ogbl-biokg",  # TODO: requires heterogeneous graphs
+    "ogbl-vessel",
+]
+OGB_GRAPH_DATASET = Literal[
+    "ogbg-molhiv",
+    "ogbg-molpcba",
+    "ogbg-ppa",
+    "ogbg-code2",
+]
+
+
+@overload
+def to_mx_array(x: None) -> None:
+    ...
+
+
+@overload
+def to_mx_array(x: np.ndarray) -> mx.array:
+    ...
+
+
+def to_mx_array(x: Union[np.ndarray, None]) -> Union[mx.array, None]:
+    if x is None:
+        return None
+    else:
+        return mx.array(x.tolist())
+
+
+class OGBDataset(Dataset):
+    def __init__(
+        self,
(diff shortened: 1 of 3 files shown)


mlx_graphs/datasets/ogb_dataset.py
```diff
@@ -0,0 +1,194 @@
...
+        i
+        for i in list(get_args(OGB_NODE_DATASET))
+        + list(get_args(OGB_EDGE_DATASET))
+        + list(get_args(OGB_GRAPH_DATASET))
+    ]
+]
+
+
+@overload
+def to_mx_array(x: None) -> None:
```
TristanBilot, line 44: what is the purpose of this overloading?

mlx_graphs/datasets/ogb_dataset.py
```diff
@@ -0,0 +1,194 @@
...
+    ):
+        self.split = split
+        super().__init__(name, base_dir)
+
+    def download(self):
+        self._load_data()
+
+    def _load_data(self):
+        if self.name in get_args(OGB_NODE_DATASET):
+            from ogb.nodeproppred import NodePropPredDataset as OGB_dataset
```
TristanBilot, line 119: are you relying on an external package? Is it required? If yes, you should add it to the dependencies.

mlx_graphs/datasets/ogb_dataset.py
```diff
@@ -0,0 +1,194 @@
...
+                    )
+                )
+            split_idx: dict = dataset.get_idx_split()  # type: ignore
+            train_idx, val_idx, test_idx = (
+                split_idx["train"],
+                split_idx["valid"],
+                split_idx["test"],
+            )
+            if self.split == "train":
+                self.graphs = [self.graphs[i] for i in train_idx]
```
TristanBilot, line 190: did you try `self.graphs = self.graphs[train_idx]`?

mlx_graphs/datasets/ogb_dataset.py
francescofarina: to avoid having typing errors/warnings in the code below where it's used (lines 149-151 and similar)

mlx_graphs/datasets/ogb_dataset.py
francescofarina: yes, it doesn't work, even if `train_idx` is converted to a list

mlx_graphs/datasets/ogb_dataset.py
francescofarina: Yes, `ogb` is required here - however, I'm importing it inside the function because it's only required to create this dataset, so only people that wants to use `OGBDataset` will have to install `ogb`. I'll add a comment in the docstring.

francescofarina: done

mlx_graphs/datasets/ogb_dataset.py
TristanBilot: hmm, strange

mlx_graphs/datasets/ogb_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+                "you can install it via pip install ogb",
+            )
+
+    def process(self):
+        try:
+            dataset = self._raw_ogb_dataset
+        except AttributeError:
+            # reload if already downloaded
+            self._load_data()
+            dataset = self._raw_ogb_dataset
```
TristanBilot, line 142: isn't it cleaner to declare `self._raw_ogb_dataset = None` in the `__init__` and check if `self._raw_ogb_dataset is None` rather than a `try/except`?

francescofarina: I didn't do that to avoid having typing errors, but yes, it's cleaner so I fixed it.

TristanBilot (approved):
🚢

PR #139, opened 2024-03-06
add ACKNOWLEDGMENTS.md

TristanBilot:


francescofarina:
@TristanBilot, in my experience this may become difficult to maintain and may become hard to use as the project grows. Isn't it better to use the `Insights` tab or the `contributors` list?

TristanBilot:
Hmm I would say that it motivates people to contribute to see their name in this file. Even if it's not totally updated everytime, I guess it's no big deal no? At least we can give it a try.

francescofarina:
Ok, let's try - worst case we'll scrap it later. Let's also add @Kheims' conversion to nx

francescofarina (approved):

Issue #123, opened 2024-03-02
Add transform to Dataset


francescofarina:
We may need/want to add a transform function to the `Dataset` class to allow for data transformations during processing. E.g. for QM9 it's common to normalize features and targets, but that currently needs to be done manually and post-data loading.

shenoynikhil:
Heyo! Awesome effort. let me try and work on this.

PR #144, opened 2024-03-08
Graph transform

shenoynikhil:
## Proposed changes

Fixes for issue #123. 

#### Description
Adds an ability to apply a callable transform to graphdata during `__getitem__`.

## Checklist


TristanBilot:
Hi @shenoynikhil, thanks for the PR! Would you mind adding some tests?

shenoynikhil:
Sorry for the delay. I added the test. It took me a bit to setup. I think the latest code in there requires `mlx==0.6.*`.

tests/datasets/test_dataset.py
```diff
@@ -80,3 +80,32 @@ def process(self):
...
+        def download(self):
+            pass
+
+        def process(self):
+            self.graphs = data_list
+
+    dataset = Dataset1(custom_transform)
+
+    for i in range(len(dataset)):
+        assert dataset[i].feat == "test_transform"
```
francescofarina, line 111: Can you delete the content of `path` after this?
We'll have to implement a fixture to automatically do that for all files created during testing now that we have multiple datasets, but we'll do that in a later PR.

francescofarina (changes requested):

tests/datasets/test_dataset.py
shenoynikhil: Added a line to remove the files at the path.

francescofarina (approved):
Looks good! I did some minor tweaking to the docs but it's great.

We may want to implement a `pre_transform` as well in a following PR to pre-transform all data before saving them @shenoynikhil, in case you're interested in looking into that.

Issue #141, opened 2024-03-06
Improve edge_weigths description in GINConv


thegodone:
Is the edge_weigths are a way to do GINE ? this is not clear in the arg definition:             edge_weights: Edge weights leveraged in message passing. Default: ``None``

Normally, we need to do first a linear projection of edge_features into node_features for GINE code (https://github.com/akensert/molgraph/blob/main/molgraph/layers/convolutional/gin_conv.py or https://github.com/aimat-lab/gcnn_keras/blob/master/kgcnn/literature/GIN/_model.py)

TristanBilot:
So basically, the `edge_weights` arg can be passed to any `MessagePassing` layer. They will just weigh the features from neighbors in the `message` function. `edge_weights` is supposed to store scalar elements for each edge, so it has a shape (E, ) or (E, 1).
Edge features are thus not implemented yet in GIN. To integrate GINE, we may want to create a dedicated `GINE` layer that applies the projection on the edge features beforehand.

thegodone:
Generally this is the second option that makes more sense cause it's just an extension of the GIN.

TristanBilot:
Would you like to work on this issue @thegodone?

thegodone:
I will but not this week

Le jeu. 7 mars 2024 à 21:30, Tristan Bilot ***@***.***> a
écrit :

PR #145, opened 2024-03-09
Add GINE capabilities to GIN

thegodone:
This is a tentative to implement GINE (adding edges)

Current code needs to define the edges size maybe this can be avoid.

## Proposed changes

tentative GINE example to address #141.

## Checklist


francescofarina:
@thegodone it would be great to have this implemented within `mlx-graphs` rather than in a notebook.

thegodone:
Of course, yes, but I am not sure of the result yet. can you first check my
code ?

Le dim. 10 mars 2024 à 10:36, Francesco Farina ***@***.***> a
écrit :

TristanBilot:
For a next PR, it would be easier if you can write your code in real `.py` files instead of notebooks, as it's complicated to leave comments on a notebook.

Here are 
- Can you add `bias=True` in the `Linear` layer you instantiate? We generally follow PyG's implementation and they use bias.
- You use `edge_projection` instead of `edge_lin_proj`...
- `and not edge_features.shape[1] == node_features.shape[1]` -> you should apply the edge projection even if the node and edge shapes match.
- `hasattr(self, 'edge_projection')` -> you can use `'edge_projection' in self` instead
- your tests do not test GINE but only GINConv

thegodone:
I update the python file gin_conv.py

mlx_graphs/nn/conv/gin_conv.py
```diff
@@ -88,13 +111,19 @@ def __call__(
...
 
         aggr_features = self.propagate(
             edge_index=edge_index,
             node_features=node_features,
-            message_kwargs={"edge_weights": edge_weights},
+            message_kwargs={"edge_weights": edge_weights,
+                            "edge_features": edge_features},
         )
+        if 'edge_projection' in self:
+            aggr_features += edge_features
```
TristanBilot: You can't sum node embeddings with edge embeddings here, you have to do that in the `message` function.

thegodone:
Thanks Tristan, for the next time I will follow PyG implementations. I would like to add the DMPNNconv what is the best way cause basically this is just MPNN but with a directed graph.

TristanBilot:
I plan to merge this PR @thegodone, however you pushed a notebook related to the Esol dataset, can you remove it and cherry pick your work to a new branch for another PR?

francescofarina:
> I plan to merge this PR @thegodone, however you pushed a notebook related to the Esol dataset, can you remove it and cherry pick your work to a new branch for another PR?

@TristanBilot I'll merge after your last review

Issue #117, opened 2024-02-29
Change in GCNConv normalization 


Kheims:
As soon as we upgrade to mlx v0.5.0 we can change the [GCNConv normalization step](https://github.com/mlx-graphs/mlx-graphs/blob/main/mlx_graphs/utils/scatter.py#L199-L219), as `mx.where` now properly handles `inf` [see](https://github.com/ml-explore/mlx/releases/tag/v0.5.0)

francescofarina:
V0.5 just came out

PR #146, opened 2024-03-10
update inv_sqrt_degree (removing workaround)

Kheims:
## Proposed changes

Removed the workaround for 'inf' values in the normalization step of the GCN  #117  as it is natively handled in mlx.
Still need to potentially change this in case boolean indexing becomes available in mlx.

## Checklist


francescofarina (approved):
Looks good!

PR #150, opened 2024-03-14
Add Elliptic bitcoin dataset

vinayhpandya:
## Added elliptic bitcoin transaction dataset 

Based on pytorch's version for elliptic bitcoin dataset
https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/datasets/elliptic.html

## Checklist

### DIFF
--- a/mlx_graphs/datasets/elliptic.py
+++ b/mlx_graphs/datasets/elliptic.py
@@ -1,8 +1,8 @@
 from typing import Optional
 
 import mlx.core as mx
-import pandas as pd
-
+import numpy as np
+import shutil
 from mlx_graphs.data import GraphData
 from mlx_graphs.datasets.dataset import Dataset
 from mlx_graphs.datasets.utils import download, extract_archive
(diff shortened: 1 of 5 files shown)


mlx_graphs/datasets/elliptic.py
```diff
@@ -0,0 +1,106 @@
...
+            "elliptic_txs_classes.csv",
+        ]
+
+    def download(self):
+        # This url is unable to download the data for elliptic bitcoin dataset
+        # lets try with pytorch geometric to verify the data
+        url = "https://data.pyg.org/datasets/elliptic/"
+        for files in self.raw_file_names:
+            download(f"{url}{files}.zip", self.raw_path)
+            extract_archive(f"{self.raw_path}/{files}.zip", self.processed_path)
```
francescofarina, line 67: why are you extracting the data in `processed_path`? I guess it should go in `raw_path`

francescofarina (changes requested):

TristanBilot:
Amazing @vinayhpandya, thanks for adding this dataset! Apart from @francescofarina's comment, it seems all good to me.

mlx_graphs/datasets/elliptic.py
vinayhpandya: Updated in the new commit. I have also added pandas in the dependency section since the tests were failing

francescofarina (commented):
Very up for discussion: looks like pandas is only used to parse the csv. Do we really need to depend on pandas? @vinayhpandya @TristanBilot wdyt?

vinayhpandya:
@francescofarina I don't think its absolutely necessary to add pandas, I can try to see if numpy can extract csv values for nodes and edges and that should work as well

francescofarina:
> @francescofarina I don't think its absolutely necessary to add pandas, I can try to see if numpy can extract csv values for nodes and edges and that should work as well

That would be great - I think it's best to not add `pandas` if `numpy` works as well!

vinayhpandya:
I tried using numpy's loadtxt functionality to manipulate the node features but it takes around 9 secs
compared to 3 secs with Pandas, is csv reader a better option?
`Time taken is 9.910717964172363
Feature shape is (203769, 167)
Time taken for pandas is 3.2206032276153564`

francescofarina:
@vinayhpandya I think the 9 seconds with numpy are ok. The dataset is processed only once and then stored in memory to be reloaded when needed, so that should be ok 👍🏻

TristanBilot:
Good catch! It would be indeed better to use numpy here, even if slower.

vinayhpandya:
@TristanBilot @francescofarina I have removed pandas and used numpy's `loadtxt` to process the files. Although this process is slightly slower than pandas (upto ~7secs)

Issue #149, opened 2024-03-12
`pre_transform` for `Dataset`


francescofarina:
**Is your feature request related to a problem? Please describe.**
#144 implemented an optional `transform` callable argument to be passed to datasets so that data can be transformed on access. Having also a pre-transform may come in handy to transform all data before saving.

**Describe the solution you'd like**
An optional `pre_transform` callable applied during `process` to all data. We'll likely need to also modify the save method or implement a `force_reload` to allow reloading of the clean dataset.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #152, opened 2024-03-17
Add Normalized features and base transforms

vinayhpandya:
Body: Added Normalized feature transform and base transforms.
1. Added unit tests for Normalize features
2. Ran pre-commit steps

This can be a potential fix for #149

## Proposed changes

To have `pre_transform/transform` as a feature step for preprocessing data before loading
Implemented Normalized features which normalizes `node_features` of graph data
 
## Checklist

### DIFF
--- a/mlx_graphs/datasets/elliptic.py
+++ b/mlx_graphs/datasets/elliptic.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Callable, Optional
 
 import mlx.core as mx
 import numpy as np
@@ -27,10 +27,15 @@ class EllipticBitcoinDataset(Dataset):
     Args:
         base_dir: Directory where to store dataset files. Default is
             in the local directory ``.mlx_graphs_data/``.
+        transforms (callable, Optional): A function/transform that
+            takes in a graphData object and returns a transformed version
+            The data object will be transformed before every access
     """
 
-    def __init__(self, base_dir: Optional[str] = None):
-        super().__init__(name="ellipticBitcoin", base_dir=base_dir)
+    def __init__(
+        self, base_dir: Optional[str] = None, transform: Optional[Callable] = None
+    ):
+        super().__init__(name="ellipticBitcoin", base_dir=base_dir, transform=transform)
 
     @property
     def raw_file_names(self):
--- a/mlx_graphs/datasets/elliptic.py
+++ b/mlx_graphs/datasets/elliptic.py
@@ -27,15 +27,27 @@ class EllipticBitcoinDataset(Dataset):
     Args:
         base_dir: Directory where to store dataset files. Default is
             in the local directory ``.mlx_graphs_data/``.
+        pre_transform (callable, Optional): A function/transform which
+            takes in a GraphData object and returns a transformed
+            version. The data will be transformed before saving to
+            the disk.
         transforms (callable, Optional): A function/transform that
             takes in a graphData object and returns a transformed version
             The data object will be transformed before every access
     """
 
     def __init__(
-        self, base_dir: Optional[str] = None, transform: Optional[Callable] = None
+        self,
+        base_dir: Optional[str] = None,
+        pre_transform: Optional[Callable] = None,
+        transform: Optional[Callable] = None,
     ):
-        super().__init__(name="ellipticBitcoin", base_dir=base_dir, transform=transform)
+        super().__init__(
+            name="ellipticBitcoin",
+            base_dir=base_dir,
+            pre_transform=pre_transform,
+            transform=transform,
+        )
 
     @property
     def raw_file_names(self):
@@ -104,4 +116,5 @@ class EllipticBitcoinDataset(Dataset):
         )
         graph.train_mask = train_mask
         graph.test_mask = test_mask
+
         self.graphs = [graph]
--- a/mlx_graphs/datasets/elliptic.py
+++ b/mlx_graphs/datasets/elliptic.py
@@ -27,11 +27,11 @@ class EllipticBitcoinDataset(Dataset):
     Args:
         base_dir: Directory where to store dataset files. Default is
             in the local directory ``.mlx_graphs_data/``.
-        pre_transform (callable, Optional): A function/transform which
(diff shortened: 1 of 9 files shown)


mlx_graphs/transforms/base_transform.py
```diff
@@ -0,0 +1,21 @@
+import copy
+from abc import ABC
+from typing import Any
+
+
+class BaseTransform(ABC):
+    r"""An abstract class for creating transformations
+
+    Transforms are a way to modify and customize
+    mlx_graphs.data.GraphData into suitable formats
```
francescofarina, line 10: mlx_graphs.data.GraphData -> `GraphData`

mlx_graphs/transforms/normalize_features.py
```diff
@@ -0,0 +1,25 @@
...
+    ) -> Union[GraphData, GraphDataBatch]:
+        if isinstance(data, GraphData):
+            for attribute in self.attributes:
+                array = getattr(data, attribute)
+                if array is not None:
+                    if array.size > 0:
+                        array = array - mx.min(array)
+                        sum_val = mx.sum(array, axis=-1, keepdims=True)
+                        new_array = array / mx.clip(sum_val, 1.0, None)
+                    setattr(data, attribute, new_array)
```
francescofarina, line 24: What is this loop normalizing? How do we make sure only node/edge/graph features are normalized and not labels or other attributes?

francescofarina (changes requested):
Great initiative @vinayhpandya.

I left a few comments

mlx_graphs/transforms/normalize_features.py
vinayhpandya: Its basically normalizing node features. I chose Elliptic bitcoin for this test (but it can work with any kind of features, node/edge). I did not find an optimized way to do this as mlx does not allow inplace division for some reason

mlx_graphs/transforms/normalize_features.py
```diff
@@ -0,0 +1,25 @@
+from typing import List, Union
+
+import mlx.core as mx
+
+from mlx_graphs.data import GraphData, GraphDataBatch
+from mlx_graphs.transforms import BaseTransform
+
+
+class NormalizeFeatures(BaseTransform):
```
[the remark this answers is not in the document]
vinayhpandya: Updated

mlx_graphs/transforms/normalize_features.py
TristanBilot: As specific attributes are passed into `attributes`, the user can easily control which attributes to normalize, so it shouldn't cause any problem regarding labels etc.

mlx_graphs/transforms/base_transform.py
```diff
@@ -0,0 +1,30 @@
...
+    in mlx_graphs.dataset.Dataset
+
+    Example:
+
+    .. code-block:: python
+
+        from mlx_graphs.datasets import Planetoid
+        from mlx_graphs.transforms import FeaturesNormalizedTransform
+
+        dataset = EllipticBitcoinDataset(transform=FeaturesNormalizedTransform())
```
TristanBilot, line 19: Shouldn't it be clearer to move this example to `FeaturesNormalizedTransform`?

TristanBilot:
Great work @vinayhpandya!

francescofarina:
@vinayhpandya I added the transforms to the docs

francescofarina (approved):

francescofarina (changes requested):
I think this type of transform should be passed as `pre_transform`. We're using the `transform` arg for transformations applied on data access (see #149)

mlx_graphs/datasets/elliptic.py
```diff
@@ -27,10 +27,15 @@ class EllipticBitcoinDataset(Dataset):
     Args:
         base_dir: Directory where to store dataset files. Default is
             in the local directory ``.mlx_graphs_data/``.
+        transforms (callable, Optional): A function/transform that
```
TristanBilot, line 30:
```suggestion
        transform: A function that
```

TristanBilot:
> I think this type of transform should be passed as `pre_transform`. We're using the `transform` arg for transformations applied on data access (see #149)

I guess the user will always index the dataset to get the graphs so the `transform` should be always called, no?

vinayhpandya:
Yes, if the transforms are present they are always called when `getitem()` is called.

francescofarina:
Yes, what I'm saying is that we can also pass it as a `pre_transform` and update the default `process` for the dataset so that all data are transformed before being saved.

vinayhpandya:
Makes sense, I will add that to `process`

mlx_graphs/transforms/base_transform.py
```diff
@@ -0,0 +1,21 @@
...
+class BaseTransform(ABC):
+    r"""An abstract class for creating transformations
+
+    Transforms are a way to modify and customize
+    ``GraphData`` by implicitly passing them as arguments
+    in ``Dataset``
+    """
+
+    def __call__(self, data: Any) -> Any:
+        return self.process(copy.copy(data))
```
francescofarina, line 15: why is the copy needed here?

mlx_graphs/transforms/base_transform.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def process(self, data: Any) -> Any:
```
francescofarina, line 17: may be worth calling this transform and make it an abstract method so that its implementation is always required (like we did for process in the `Dataset` class)

francescofarina (changes requested):
This looks great to me! Great work @vinayhpandya - just two minor questions before approving and merging

mlx_graphs/transforms/base_transform.py
vinayhpandya: So that we don't accidently change the values inplace

francescofarina: ok, makes sense for when it's passed to `transform` instead of `pre_transform` 👍

francescofarina (approved):
🚢

Issue #110, opened 2024-02-27
Convert GraphData from/to NetworkX with features


TristanBilot:
### Step 1
`from_networkx` has been implemented in https://github.com/mlx-graphs/mlx-graphs/pull/109, but it lacks the cases where attributes/features exist in the `GraphData`.

One may want to follow the same methodology as in [PyG](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/utils/convert.html#from_networkx).

### Step 2
The same applies for `to_networkx`. This function is not yet implemented.

Kheims:
I can handle this but not until end of week as i am fully dedicated to neighbor sampler as of now

TristanBilot:
It's ok, neighbor sampling is more important 😅. Anyone interested can go with this issue.

francescofarina:
Completed by @vinayhpandya in #154.

PR #154, opened 2024-03-24
Add utils to convert networkx to GraphData

vinayhpandya:
1. Updated to_networkx function to add node, graph and edge features as well as labels
2. Added from_networkx utility to convert networkx.graph object to GraphData.

## Proposed changes

Fix for #110 

## Checklist

### DIFF
--- a/mlx_graphs/utils/convert.py
+++ b/mlx_graphs/utils/convert.py
@@ -1,4 +1,7 @@
+from collections import defaultdict
 from typing import Any
+
+import mlx.core as mx
 
 from mlx_graphs.data import GraphData
 
@@ -34,15 +37,63 @@ def to_networkx(
 
     G = nx.DiGraph()
 
+    if data.graph_features:
+        G.graph["graph_features"] = data.graph_features.tolist()
+
+    if data.graph_labels:
+        G.graph["graph_labels"] = data.graph_labels.tolist()
+
     if data.num_nodes is None:
         return G
 
-    G.add_nodes_from(range(data.num_nodes))
+    # G.add_nodes_from(range(data.num_nodes))
 
-    for v, w in data.edge_index.T.tolist():
+    node_features_present = data.node_features is not None
+    node_labels_present = data.node_labels is not None
+
+    node_attrs = {}
+    for i in range(data.num_nodes):
+        if node_features_present:
+            node_attrs["features"] = data.node_features[i].tolist()
+        if node_labels_present:
+            node_attrs["label"] = data.node_labels[i].item()
+        G.add_node(i, **node_attrs)
+
+    edge_features_present = data.edge_features is not None
+    edge_attrs = {}
+    for i, (v, w) in enumerate(data.edge_index.T.tolist()):
         if remove_self_loops and v == w:
             continue
-
-        G.add_edge(v, w)
+        if edge_features_present:
+            edge_attrs["features"] = data.edge_features[i].tolist()
+        G.add_edge(v, w, **edge_attrs)
 
     return G
+
+
+def from_networkx(data: Any) -> GraphData:
+    data_dict = defaultdict(list)
+
+    edge_index = mx.array(list(data.edges())).T
+    data_dict["edge_index"] = edge_index
+
+    for attr in ["features", "label"]:
+        for entity in ["edge", "node", "graph"]:
+            key = f"{entity}_{attr}"
+            values = []
+
+            if entity == "edge":
+                for _, _, feat_dict in data.edges(data=True):
+                    if attr in feat_dict:
+                        values.append(feat_dict[attr])
+            elif entity == "node":
+                for _, feat_dict in data.nodes(data=True):
+                    if attr in feat_dict:
+                        values.append(feat_dict[attr])
+            elif entity == "graph" and attr in data.graph:
+                values = [data.graph[attr]]
+
+            if values:
+                data_dict[key] = mx.array(values)
+
+    return GraphData(**data_dict)
--- a/mlx_graphs/utils/convert.py
+++ b/mlx_graphs/utils/convert.py
@@ -59,7 +59,7 @@ def to_networkx(
     for i in range(data.num_nodes):
         if data.node_features is not None:
             node_attrs["features"] = data.node_features[i].tolist()
-        if data.node_labels:
+        if data.node_labels is not None:
             node_attrs["label"] = data.node_labels[i].item()
         G.add_node(i, **node_attrs)
 
--- a/mlx_graphs/utils/convert.py
+++ b/mlx_graphs/utils/convert.py
@@ -1,7 +1,8 @@
 from collections import defaultdict
-from typing import Any
+from typing import Any, Union
 
 import mlx.core as mx
+import networkx as nx
 
 from mlx_graphs.data import GraphData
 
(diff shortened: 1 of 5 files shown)


mlx_graphs/utils/convert.py
```diff
@@ -34,15 +37,63 @@ def to_networkx(
...
-
-        G.add_edge(v, w)
+        if edge_features_present:
+            edge_attrs["features"] = data.edge_features[i].tolist()
+        G.add_edge(v, w, **edge_attrs)
 
     return G
+
+
+def from_networkx(data: Any) -> GraphData:
```
francescofarina, line 74: Can you use the proper networkx types instead of `Any` here?

francescofarina, line 74: Can you add the docstrings?

mlx_graphs/utils/convert.py
```diff
@@ -34,15 +37,63 @@ def to_networkx(
...
+    data_dict = defaultdict(list)
+
```
francescofarina, line 76: You may need to import networkx in here

pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
     "mlx==0.7.*; sys_platform == 'darwin'",
     "numpy==1.26.3",
     "requests==2.31.0",
+    "networkx",
```
francescofarina, line 23: we may not want to have this - people wanting to use the conversion functionalities will have to install it, but it shouldn't be a dependency of ours. We can add it to the optional dependencies.

francescofarina (changes requested):

mlx_graphs/utils/convert.py
TristanBilot: What we usually do is to provide a message to the user if the package is not installed (see the try/except block [here](mlx_graphs/datasets/lanl_dataset.py))

mlx_graphs/utils/convert.py
```diff
@@ -34,15 +37,63 @@ def to_networkx(
 
     G = nx.DiGraph()
 
+    if data.graph_features:
```
TristanBilot, line 40:
```suggestion
    if data.graph_features is not None:
```

mlx_graphs/utils/convert.py
```diff
@@ -34,15 +37,63 @@ def to_networkx(
...
+        G.graph["graph_features"] = data.graph_features.tolist()
+
+    if data.graph_labels:
```
TristanBilot, line 43:
```suggestion
    if data.graph_labels is not None::
```

pyproject.toml
TristanBilot: I think we can add networkx to the test dependencies?
Or we have to use `@pytest.mark.skipif(not networkx_installed, reason="networkx is not installed")` like for `test_to_networkx()`.

tests/utils/test_convert.py
```diff
@@ -39,3 +50,18 @@ def test_to_networkx():
...
+
+
+def test_from_networkx():
+    karate_club_dataset = KarateClubDataset()
+    karate_club_networkx_graph = to_networkx(karate_club_dataset.graphs[0])
+    restored_karate_club_dataset = from_networkx(karate_club_networkx_graph)
+
+    assert mx.array_equal(
+        restored_karate_club_dataset.node_features,
+        karate_club_dataset.graphs[0].node_features,
```
TristanBilot, line 62:
```suggestion
        karate_club_dataset[0].node_features,
```

tests/utils/test_convert.py
```diff
@@ -39,3 +50,18 @@ def test_to_networkx():
...
+    ), "Node features are not equal"
+    assert (
+        restored_karate_club_dataset.edge_index.shape[1]
+        == karate_club_dataset.graphs[0].edge_index.shape[1]
```
TristanBilot, line 66:
```suggestion
        == karate_club_dataset[0].edge_index.shape[1]
```

TristanBilot (changes requested):
Thanks for the great work @vinayhpandya!
Just left few comments. If you have some time, could you add some extra tests to verify networkx functions like `networkx_graph.number_of_nodes()` (see existing tests in `test_to_networkx()`).

Don't forget to add you function to `convert.rst` so that it is visible in the doc ;)

mlx_graphs/utils/convert.py
```diff
@@ -30,19 +34,95 @@ def to_networkx(
...
-
-        G.add_edge(v, w)
+        if edge_features_present:
+            edge_attrs["features"] = data.edge_features[i].tolist()
+        G.add_edge(v, w, **edge_attrs)
 
     return G
+
+
+def from_networkx(data: Union[nx.Graph]) -> GraphData:
```
francescofarina, line 74: `Union[nx.Graph]` -> `Union[nx.Graph, nx.DiGraph]`?

mlx_graphs/utils/convert.py
```diff
@@ -30,19 +34,95 @@ def to_networkx(
...
+    """Converts a :obj:`networkx.Graph` or :obj:`networkx.DiGraph` to a
+    :class:`mlx_graphs.data.GraphData` instance.
+
+    Args:
+        data (Union[nx.Graph]): A networkx graph
```
francescofarina, line 79: No need to specify the type here as it's inferred from the signature

pyproject.toml
```diff
@@ -31,6 +31,7 @@ dev = [
 test = [
   "pytest==7.4.4",
   "scipy==1.12.0",
+  "networkx",
```
francescofarina, line 34: add version

mlx_graphs/utils/convert.py
```diff
@@ -1,4 +1,8 @@
-from typing import Any
+from collections import defaultdict
+from typing import Any, Union
+
+import mlx.core as mx
+import networkx as nx
```
francescofarina, line 5: this requires networkx as a main dependency (which I'm still not sure we want to have)

francescofarina (changes requested):

mlx_graphs/utils/convert.py
vinayhpandya: nx.Digraph inherits from graph so I guess we this can cover both cases
https://networkx.org/documentation/networkx-1.10/_modules/networkx/classes/digraph.html#DiGraph

mlx_graphs/utils/convert.py
vinayhpandya: Is there a way to not add it and still have `Union[nx.Graph]` as input? I tried to remove it but it keeps giving me the error related to dependency

mlx_graphs/utils/convert.py
francescofarina: Ok, then I guess the `Union` is not needed

mlx_graphs/utils/convert.py
francescofarina: You can wrap the import in a `try/except`

```python
try:
    import networkx as nx
except ImportError:
    raise ImportError("networkx is required to convert from nextworkx graphs - run `pip install networkx`")
```

francescofarina: I'll check whether that causes issues when compiling the documentation

mlx_graphs/utils/convert.py
vinayhpandya: Can I keep it `Any` here and then try to import `networkx` that way the user will not have to install networkx for every function in the future

francescofarina: No, should we need to, we'll make networkx a dependency, but I'd rather not use `Any`

mlx_graphs/utils/convert.py
vinayhpandya: I think I need to import networkx before as we want a hard dependency on input
`from_networkx` needs to have `nx.Graph` as input and that requires networkx to be imported before

francescofarina: Yes, you can do it in a try/except in the imports in the preamble.

francescofarina:
@vinayhpandya I fixed the import and made sure the documentation will compile. Give it a try to see if it's ok.

francescofarina (approved):

francescofarina:
@TristanBilot I'll merge if you don't have any other comments

Issue #128, opened 2024-03-03
ESOL dataset using directly RDkit descriptors


thegodone:
Can you provide an example from scratch off to convert Molecules objects into Atom / Nodes and Bond / Edge features for ESOL dataset ?

[ ] using RDkit
[ ] for GIN / GINE / DMPNN (chemProp) / AttentiveFP

francescofarina:
@thegodone we currently have a basic example of how to build custom datasets [here](https://mlx-graphs.github.io/mlx-graphs/tutorials/quickstart.html).

Having a specific tutorial on how to use RDKit to convert the ESOL dataset and testing it with different architectures is definitely interesting but I think it's probably better suited to live outside of the main `mlx-graphs` repository, perhaps as a tutorial or a standalone project.

francescofarina:
It would be great to have the ESOL dataset built within the library and we may as well do that over the coming weeks (or feel free to contribute to it 😉) but I think your question was more around showing how to make the conversion.

thegodone:
I add this PR to make a demo on how to integrate a molecular dataset from scratch using rdkit and loading a file containing smiles and a target. I still have an issue cause the batch and the default dataset informations are not available (python errors)

PR #155, opened 2024-03-24
Esol

thegodone:
## Proposed changes

Current status:

I try to set up a custom dataset based on so called  "ESOL / SAMPL / delaney" dataset, I add rkdit to generate basic node and edge features and I created the graphData for each molecules. And finally the dataset. But I get an error when I when to get defaults infos on the dataset or generate the batches.  

Goal:

Generally lot of DS are working on private data and those data are not standardized:
In this demo, the goal is to access and use any type of data : any file (locally or remotely) and any field (smiles , targets).
see #128 

## Checklist

### DIFF
--- a/mlx_graphs/nn/conv/gin_conv.py
+++ b/mlx_graphs/nn/conv/gin_conv.py
@@ -3,6 +3,7 @@ from typing import Optional
 import mlx.core as mx
 import mlx.nn as nn
 
+from mlx_graphs.nn.linear import Linear
 from mlx_graphs.nn.message_passing import MessagePassing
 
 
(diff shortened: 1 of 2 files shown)


TristanBilot:
@thegodone, I just looked at your error. It seems that there is an empty `node_labels` in the graphs from your dataset.
You have to properly load the array with the labels.

```python
dataset[0]
>>> GraphData(
	edge_index(shape=(2, 68), uint32)
	node_features(shape=(32, 49), float32)
	edge_features(shape=(34, 4), int32)
	node_labels(shape=(), float32))  # the empty array causes the error
```

francescofarina:
@thegodone I'm not sure I understand

> ## I find this workaround label = mx.ones((1, ))*row.LogS
> that fix the next cell but not the last one.
> Dataset attributes
> ## Number of graphs: 1127
> AttributeError Traceback (most recent call last) Cell In[22], line 5 3 print("-" * 20) 4 print(f"Number of graphs: {len(dataset)}") ----> 5 print(f"Number of node features: {dataset.num_node_features}") 6 print(f"Number of edge features: {dataset.num_edge_features}") 7 print(f"Number of graph features: {dataset.num_graph_features}")
> 
> AttributeError: 'list' object has no attribute 'num_node_features'

@thegodone I'm not sure I understand this. Can you provide a reproducible example of the issue you're encountering?

PR #157, opened 2024-03-26
fix esol GIN case

thegodone:
This fix the example of GIN with ESOL dataset from scratch using rdkit backend for features computation.

## Proposed changes

I create an example of GIN (3 layers) for ESOL dataset. This is for demo proposal since we don't have batch or layernorm.

## Checklist

### DIFF
--- a/docs/source/tutorials/examples/ESOL_mlx_graph_demo.ipynb
+++ b/docs/source/tutorials/examples/ESOL_mlx_graph_demo.ipynb
@@ -1,0 +1,731 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "1de1dc48-49a2-4931-9703-1afd3eda656c",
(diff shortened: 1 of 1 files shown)


thegodone:
I understand now your point of normalization layers.

francescofarina:
Thanks this tutorial @thegodone - if we want to merge and publish this in the documentation, it probably needs a bit of extra work to make it more didactic (e.g., explanation of the problem, step-by-step guidance, ...). What do you think?

thegodone:
Sure I will add SbS guidance and also plots.
I also think it would be nice to have true CV code anyway.
This is still in progress.
I've just made it works with new GINE version.
[image: image.png]

Le mar. 26 mars 2024 à 08:46, Francesco Farina ***@***.***> a
écrit :

thegodone:
Ok the GINE version is not working yet.

PR #159, opened 2024-03-27
Additional utils ( for isolated nodes and self loops)

Kheims:
## Proposed changes

Added some graph utilities 

- has_isolated_nodes
- mask_isolated_nodes (return mask) so that user can filter out isolated nodes from node features 
- has_self_loops

## Checklist

### DIFF
--- a/mlx_graphs/data/data.py
+++ b/mlx_graphs/data/data.py
@@ -3,7 +3,7 @@ from typing import Literal, Optional, Union
 import mlx.core as mx
 import numpy as np
 
-from mlx_graphs.utils import remove_self_loops
+from mlx_graphs.utils import has_isolated_nodes, has_self_loops
 
 
 class GraphData:
@@ -179,10 +179,8 @@ class GraphData:
     def has_isolated_nodes(self) -> bool:
         """Returns a boolean of whether the graph has isolated nodes or not.
         (ie: nodes that does not have an edge with any other nodes )"""
-        edge_index, num_nodes = self.edge_index, self.num_nodes
-        edge_index = remove_self_loops(edge_index)
-        return np.unique(edge_index.reshape(-1)).size < num_nodes
+        return has_isolated_nodes(self.edge_index, self.num_nodes)
 
     def has_self_loops(self) -> bool:
         """Returns a boolean of whether the graph contains self loops."""
-        return ((self.edge_index[0] == self.edge_index[1]).sum() > 0).item()
+        return has_self_loops(self.edge_index)
--- a/mlx_graphs/data/data.py
+++ b/mlx_graphs/data/data.py
@@ -3,6 +3,8 @@ from typing import Literal, Optional, Union
 import mlx.core as mx
 import numpy as np
 
+from mlx_graphs.utils import remove_self_loops
+
 
 class GraphData:
     """
@@ -173,3 +175,12 @@ class GraphData:
         if "index" in key:
             return self.num_nodes
         return None
+
+    def has_isolated_nodes(self) -> bool:
+        """Returns a boolean of wether the graph has isolated nodes or not"""
+        edge_index, num_nodes = self.edge_index, self.num_nodes
+        edge_index = remove_self_loops(edge_index)
+        return np.unique(edge_index.reshape(-1)).size < num_nodes
+
+    def has_self_loops(self) -> bool:
+        return (self.edge_index[0] == self.edge_index[1]).sum().item() > 0
--- a/mlx_graphs/data/data.py
+++ b/mlx_graphs/data/data.py
@@ -184,5 +184,5 @@ class GraphData:
         return np.unique(edge_index.reshape(-1)).size < num_nodes
 
     def has_self_loops(self) -> bool:
-        """Returns a boolean of whether the graph contains self loops"""
+        """Returns a boolean of whether the graph contains self loops."""
         return ((self.edge_index[0] == self.edge_index[1]).sum() > 0).item()
--- a/mlx_graphs/data/data.py
+++ b/mlx_graphs/data/data.py
@@ -177,8 +177,8 @@ class GraphData:
         return None
 
     def has_isolated_nodes(self) -> bool:
-        """Returns a boolean of whether the graph has isolated nodes or not.
-        (ie: nodes that does not have an edge with any other nodes )"""
+        """Returns a boolean of whether the graph has isolated nodes or not
+        (i.e., nodes that don't have a link to any other nodes)"""
         return has_isolated_nodes(self.edge_index, self.num_nodes)
 
     def has_self_loops(self) -> bool:
--- a/mlx_graphs/data/data.py
+++ b/mlx_graphs/data/data.py
@@ -177,10 +177,12 @@ class GraphData:
         return None
 
     def has_isolated_nodes(self) -> bool:
-        """Returns a boolean of wether the graph has isolated nodes or not"""
+        """Returns a boolean of whether the graph has isolated nodes or not.
+        (ie: nodes that does not have an edge with any other nodes )"""
         edge_index, num_nodes = self.edge_index, self.num_nodes
         edge_index = remove_self_loops(edge_index)
         return np.unique(edge_index.reshape(-1)).size < num_nodes
 
     def has_self_loops(self) -> bool:
-        return (self.edge_index[0] == self.edge_index[1]).sum().item() > 0
+        """Returns a boolean of whether the graph contains self loops"""
+        return ((self.edge_index[0] == self.edge_index[1]).sum() > 0).item()
--- a/mlx_graphs/utils/transformations.py
+++ b/mlx_graphs/utils/transformations.py
@@ -456,3 +456,17 @@ def mask_isolated_nodes(
         return mx.array(indices)
 
     return mx.array(np.setdiff1d(np.arange(num_nodes), indices))
+
+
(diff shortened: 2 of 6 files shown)


mlx_graphs/data/data.py
```diff
@@ -173,3 +175,12 @@ def __inc__(self, key: str, *args, **kwargs) -> Union[int, None]:
         if "index" in key:
             return self.num_nodes
         return None
+
+    def has_isolated_nodes(self) -> bool:
+        """Returns a boolean of wether the graph has isolated nodes or not"""
```
TristanBilot, line 180:
```suggestion
        """Returns a boolean of whether the graph has isolated nodes or not."""
```

mlx_graphs/data/data.py
```diff
@@ -173,3 +175,12 @@ def __inc__(self, key: str, *args, **kwargs) -> Union[int, None]:
...
+        edge_index, num_nodes = self.edge_index, self.num_nodes
+        edge_index = remove_self_loops(edge_index)
+        return np.unique(edge_index.reshape(-1)).size < num_nodes
+
+    def has_self_loops(self) -> bool:
```
francescofarina, line 185: it's obvious, but let's have a docstring here :)

francescofarina (changes requested):
Looks good! Just one small comment and a question: shall we make `has_isolated_nodes` and `has_self_loops` properties rather than methods?

mlx_graphs/data/data.py
TristanBilot, line 180: Can you better explain what defines an isolated node in the docstring please?

mlx_graphs/data/data.py
TristanBilot, line 185: missing docstring

mlx_graphs/data/data.py
```diff
@@ -173,3 +175,12 @@ def __inc__(self, key: str, *args, **kwargs) -> Union[int, None]:
...
+        return (self.edge_index[0] == self.edge_index[1]).sum().item() > 0
```
TristanBilot, line 186:
```suggestion
        return ((self.edge_index[0] == self.edge_index[1]).sum() > 0).item()
```

mlx_graphs/utils/transformations.py
```diff
@@ -424,3 +424,35 @@ def coalesce(edge_index: mx.array) -> mx.array:
...
+        num_nodes : number of nodes of the graph.
+        filter_isolated : wether to filter isolated or non isolated nodes.
+
+    Returns:
+        a boolean mask of size ``num_nodes`` where ``True`` means the node isn't
+        isolated and ``false``means it is.
+
+    Example :
+        >>> edge_index = mx.array([[0, 2, 0], [2, 0, 0]])
+        >>> mask = remove_isolated_nodes(edge_index, 3)
```
TristanBilot, line 446:
```suggestion
        >>> mask = mask_isolated_nodes(edge_index, 3)
```

mlx_graphs/utils/transformations.py
```diff
@@ -424,3 +424,35 @@ def coalesce(edge_index: mx.array) -> mx.array:
...
+        >>> mask
+        mx.array([0,2])
+        >>> mask = remove_isolated_nodes(edge_index, 3, filter_isolated=False)
```
TristanBilot, line 449:
```suggestion
        >>> mask = mask_isolated_nodes(edge_index, 3, filter_isolated=False)
```

tests/utils/test_transformations.py
```diff
@@ -208,3 +209,33 @@ def test_remove_duplicated_directed_edges():
...
+    expected_mask = mx.array([1], dtype=mx.uint32)
+    assert mx.array_equal(mask, expected_mask), "mask_isolated_nodes failed"
+
+    edge_index = mx.array([[0, 1, 0], [1, 0, 0]])
+    mask = mask_isolated_nodes(edge_index, 3, filter_isolated=False)
+    expected_mask = mx.array([2], dtype=mx.uint32)
+    assert mx.array_equal(mask, expected_mask), "mask_isolated_nodes failed"
+
+
+test_mask_isolated_nodes()
```
TristanBilot: rm

mlx_graphs/utils/transformations.py
```diff
@@ -424,3 +424,35 @@ def coalesce(edge_index: mx.array) -> mx.array:
...
+    """Returns a mask with isolated nodes set to ``False`` to filter them out if needed.
+
+    Args:
+        edge_index : edge index from which to remove isolated nodes.
+        num_nodes : number of nodes of the graph.
+        filter_isolated : wether to filter isolated or non isolated nodes.
+
+    Returns:
+        a boolean mask of size ``num_nodes`` where ``True`` means the node isn't
+        isolated and ``false``means it is.
```
TristanBilot, line 442:
```suggestion
        isolated and ``False``means it is.
```

mlx_graphs/utils/transformations.py
```diff
@@ -424,3 +424,35 @@ def coalesce(edge_index: mx.array) -> mx.array:
...
+) -> mx.array:
+    """Returns a mask with isolated nodes set to ``False`` to filter them out if needed.
+
+    Args:
+        edge_index : edge index from which to remove isolated nodes.
+        num_nodes : number of nodes of the graph.
+        filter_isolated : wether to filter isolated or non isolated nodes.
+
+    Returns:
+        a boolean mask of size ``num_nodes`` where ``True`` means the node isn't
```
TristanBilot, line 441: please begin all your sentences with an uppercase :)

TristanBilot (changes requested):
Thanks @Kheims, these functions are useful. 
Could you implement `has_isolated_nodes` also as a standalone function (like for `mask_isolated_nodes`), so that one can use this function not necessary from a `GraphData`  object.

Kheims:
> Looks good! Just one small comment and a question: shall we make `has_isolated_nodes` and `has_self_loops` properties rather than methods?

Given that we don't encourage mutability of a graph, i think that having them as properties would be better, wdyt ?

TristanBilot:
> Looks good! Just one small comment and a question: shall we make `has_isolated_nodes` and `has_self_loops` properties rather than methods?

On this, I would say it's more intuitive to stay with functions, as we expect these functions to take some time to compute.

mlx_graphs/data/data.py
```diff
@@ -173,3 +175,14 @@ def __inc__(self, key: str, *args, **kwargs) -> Union[int, None]:
         if "index" in key:
             return self.num_nodes
         return None
+
+    def has_isolated_nodes(self) -> bool:
+        """Returns a boolean of whether the graph has isolated nodes or not.
+        (ie: nodes that does not have an edge with any other nodes )"""
+        edge_index, num_nodes = self.edge_index, self.num_nodes
+        edge_index = remove_self_loops(edge_index)
+        return np.unique(edge_index.reshape(-1)).size < num_nodes
```
francescofarina, line 184: I think now you can call the util function in here instead of rewriting it

mlx_graphs/utils/transformations.py
```diff
@@ -424,3 +424,49 @@ def coalesce(edge_index: mx.array) -> mx.array:
     edge_index = mx.stack([col, row])
 
     return edge_index
+
+
+@validate_edge_index
+def mask_isolated_nodes(
+    edge_index: mx.array, num_nodes: int, filter_isolated: bool = True
+) -> mx.array:
+    """Returns a mask with isolated nodes set to ``False`` to filter them out if needed.
```
francescofarina, line 433: @Kheims I'm not sure whether `mask_isolated_nodes` is the best name here - basically with `filter_isolated` you can change the behavior to filter the non_isolated ones. Or maybe we can rename `filter_isolated`? wdyt?

Kheims: I don't like mask_isolated_nodes, wdyt about get_isolated_nodes_indices for the function ? or (return_isolated / isolated_only ) for the parameter ?

mlx_graphs/utils/transformations.py
```diff
@@ -456,3 +456,17 @@ def mask_isolated_nodes(
...
+    """Returns a boolean of whether the graph has isolated nodes or not.
+    (ie: nodes that does not have an edge with any other nodes )"""
+    edge_index = remove_self_loops(edge_index)
+    return np.unique(edge_index.reshape(-1)).size < num_nodes
+
+
+@validate_edge_index
+def has_self_loops(edge_index: mx.array) -> bool:
+    """Returns a boolean of whether the graph contains self loops"""
+    return ((edge_index[0] == edge_index[1]).sum() > 0).item()
```
TristanBilot, line 472: Would be nice to have a comprehensive docstring with args, returns, etc, as for `mask_isolated_nodes`. And maybe one simple example.

TristanBilot (changes requested):
Don't forget to add your functions to the docs plz 📝

mlx_graphs/utils/transformations.py
TristanBilot: it's not returning indices but a boolean mask here.

francescofarina: `get_isolated_nodes_mask` makes sense, but I'd rename `filter_isolated` to `complement` as it's basically returning the isolated mask of the complement when set to false

francescofarina (approved):
🚢

PR #161, opened 2024-04-03
add is_directed and is_undirected to GraphData

### LINKED ISSUE #114: Add some util methods to GraphData
We would enjoy to run some graph util methods directly from the `GraphData` object.

**Ex:**
```python
g = GraphData(...)
>>> g.is_undirected()
>>> g.is_directed()
>>> g.has_self_loops()
>>> g.has_isolated_nodes()
```

Some of these functions are already implemented in mlx-graphs and can be simply called, like `is_undirected` and `is_directed`. Others need to be implemented.

francescofarina:
## Proposed changes

Adding `is_directed` and `is_undirected` methods to `GraphData`. Closes #114.

## Checklist

### DIFF
--- a/mlx_graphs/data/data.py
+++ b/mlx_graphs/data/data.py
@@ -4,6 +4,7 @@ import mlx.core as mx
 import numpy as np
 
 from mlx_graphs.utils import has_isolated_nodes, has_self_loops
+from mlx_graphs.utils.topology import is_undirected
 
 
 class GraphData:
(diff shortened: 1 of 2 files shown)


TristanBilot (approved):

TristanBilot:
Shall we add few tests?

francescofarina:
the methods are already tested [here](https://github.com/mlx-graphs/mlx-graphs/blob/main/tests/utils/test_topology.py)

TristanBilot:
Yeah, I meant to test the method in the GraphData object itself, just to ensure the class has this method, but if you don't think this is mandatory let's not add it.

francescofarina:
> Yeah, I meant to test the method in the GraphData object itself, just to ensure the class has this method, but if you don't think this is mandatory let's not add it.

Yes, done!

TristanBilot:
Good!

PR #162, opened 2024-04-03
make edge_index required in GraphData

francescofarina:
## Proposed changes

This makes `edge_index` a required arg when instantiating a `GraphData`. It allows not having to check whether the edge_index is not None within the framework. I think it's fair to assume that having and `edge_index` is the minimum requirement for a `GraphData`.

## Checklist


TristanBilot (approved):
🚢

PR #165, opened 2024-04-23
Rudimentry benchmarks for mlx and torch_cluster

vinayhpandya:
Added some benchmarks for random walks which were generated by mlx, torch and numpy

## Proposed changes

Added some benchmarks to compare torch_cluster, mlx and numpy

## Checklist

### DIFF
--- a/benchmarks/framework_comparison/random_walks/launcher.py
+++ b/benchmarks/framework_comparison/random_walks/launcher.py
@@ -1,0 +1,112 @@
+import platform
+import timeit
+from importlib.metadata import version
+
+import mlx.core as mx
+import numpy as np
+import torch
+import torch_geometric.datasets as pyg_datasets
+from setup_mlx import mlx_random_walks
+from setup_numpy import random_walks_numpy
+from setup_torch_cluster import torch_cluster_random_walk
+from torch.utils.data import DataLoader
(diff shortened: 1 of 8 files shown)


vinayhpandya:
Maybe, currently since mlx arrays do not allow indexing, the fastest way I could think is converting the row pointers to numpy and then calculating the random walks.

benchmarks/framework_comparison/random_walks/launcher.py
```diff
@@ -0,0 +1,92 @@
...
+
+        times = timeit.Timer(
+            lambda: benchmark(
+                framework,
+                edge_index=edge_index,
+                start_indices=start_indices,
+                walk_length=10000,
+                compile=compile,
+            )
+        ).repeat(repeat=1, number=1)
```
francescofarina, line 70: This generates just one me sample, you may want to increase the `repeat` number to get the best out of (say) 5 runs

benchmarks/framework_comparison/random_walks/results.md
```diff
@@ -0,0 +1,7 @@
+
```
TristanBilot, line 1: Can you please add your setup in the file before your results (M chip, mlx/mlx-graphs/pyg/numpy versions)

PR #166, opened 2024-05-09
Hetero data

vinayhpandya:
## Proposed changes

Added HeteroGraphData class for enabling usage of Heterogeneous graphs

## Checklist


mlx_graphs/data/heteroData.py
```diff
@@ -0,0 +1,220 @@
...
+
+        prefix = "\n\t"
+        return f"{type(self).__name__}({prefix + prefix.join(strings)})"
+
+    def to_dict(self) -> dict:
+        """Converts the Data object to a dictionary.
+
+        Returns:
+            A dictionary with all attributes of the `GraphData` object.
+        """
```
francescofarina, line 56: `HeteroGraphData`

mlx_graphs/data/heteroData.py
```diff
@@ -0,0 +1,220 @@
...
+            else None
+        )
+        return is_undirected(edge_index, edge_features)
+
+    def is_directed(self, edge_type: str) -> bool:
+        """
+        Returns a boolean of whether the given
+        edge type is directed or not.
+        """
+        return not self.is_undirected(edge_type)
```
francescofarina, line 220: Shall these methods exist also for the whole graph? E.g. `is_undirected()` will return `True` if, for all the `edge_type`, the graph is undirected.

mlx_graphs/datasets/dataset.py
```diff
@@ -166,11 +166,20 @@ def _load(self):
...
+                if labels is not None:
+                    flattened_labels.append(labels)
+            elif isinstance(g, HeteroGraphData):
+                if task == "node":
+                    labels = [v for v in g.collect(f"{task}_labels", True).values()]
+                else:
+                    labels = [getattr(g, f"{task}_labels", None)]
+                flattened_labels.extend(
+                    [label for label in labels if label is not None]
+                )
```
francescofarina, line 182: Would it make more sense to return a dict with the classes per node/edge type?

francescofarina (changes requested):
Can you add a test also for a dummy dataset? That way we can figure out whether we'd need a separate class to handle datasets with heterogenous data.

tests/data/test_hetero_data.py
```diff
@@ -0,0 +1,39 @@
...
+    expected_num_edges_dict = {
+        ("author", "writes", "paper"): 2,
+        ("paper", "cites", "paper"): 2,
+    }
+    assert hetero_data.num_edges == expected_num_edges_dict
+    expected_num_edge_classes_dict = {
+        ("author", "writes", "paper"): 2,
+        ("paper", "cites", "paper"): 2,
+    }
+    assert hetero_data.num_edge_classes == expected_num_edge_classes_dict
```
TristanBilot, line 39: tests are missing here: `num_node_features`, `num_edge_features`, `num_node_classes`, `num_nodes`, `to_dict`, etc.

mlx_graphs/data/data.py
```diff
@@ -188,3 +188,355 @@ def is_undirected(self) -> bool:
...
+
+        Returns:
+            The dimension where concatenation will happen when batching.
+        """
+        if "index" in key:
+            return 1
+        return 0
+
+    def __inc__(self, key: str, type_key: str, *args, **kwargs) -> Union[int, None]:
+        """
```
TristanBilot, line 480: only add `__inc__` and `__cat_dim__` if `HeteroGraphData` supports batching.
I don't see any tests with batching though, so you may want to either test batching with hetero graphs or remove that from the current PR and indicate the user that batching does not exist yet for `HeteroGraphData`

tests/data/test_hetero_data.py
TristanBilot: You can take the same tests from `GraphData` for example.

mlx_graphs/datasets/dataset.py
```diff
@@ -166,11 +166,20 @@ def _load(self):
...
-            for g in self.graphs
-            if getattr(g, f"{task}_labels") is not None
-        ]
+        flattened_labels = []
+        for g in self.graphs:
+            if isinstance(g, GraphData):
+                labels = getattr(g, f"{task}_labels")
+                if labels is not None:
+                    flattened_labels.append(labels)
+            elif isinstance(g, HeteroGraphData):
```
TristanBilot, line 175: would be also nice to have a test for this case within `test_dataset.py`

TristanBilot (changes requested):
I left few comments, mostly about missing tests, but the overall code seems great to me! Good work @vinayhpandya :)

TristanBilot (approved):

mlx_graphs/datasets/dataset.py
vinayhpandya: Wouldn't this also change the way we handle `GraphData`?

francescofarina: yes, that's why I was a bit concerned about having the same dataset class for `GraphData` and `HeteroGraphData`. I guess we can keep it as is for now and change it later if needed.

mlx_graphs/datasets/dataset.py
```diff
@@ -2,7 +2,7 @@
 import os
 import pickle
 from abc import ABC, abstractmethod
-from typing import Callable, Literal, Optional, Sequence, Union
+from typing import Callable, Dict, Literal, Optional, Sequence, Union
```
francescofarina, line 5: @vinayhpandya there's no need to use `Dict` from typing, you can use the standard `dict` - I'll change that

francescofarina (approved):

PR #167, opened 2024-06-23
Add IMDB and DBLP dataset

vinayhpandya:
1. Add tests
2. Add documentation
3. Add 2 more datasets

## Proposed changes

Added IMDB dataset which consist of
movies (4,278 nodes), 
actors (5,257 nodes), 
and directors (2,081 nodes)

## Checklist


mlx_graphs/datasets/IMDB.py
```diff
@@ -0,0 +1,94 @@
...
+            "labels.npy",
+            "train_val_test_idx.npz",
+        ]
+
+    def download(self):
+        url = "https://www.dropbox.com/s/g0btk9ctr1es39x/IMDB_processed.zip?dl=1"
+        path = download(url=url, path=self.raw_path)
+        new_path = path.split("?")[-2]
+        os.rename(path, new_path)
+        print("path is ", path)
```
francescofarina, line 49: rm?

mlx_graphs/datasets/IMDB.py
francescofarina, line 1: please use a lowercase name for the file (`imdb.py`)

francescofarina (changes requested):
Looks great!

Instead of importing `spicy` can you do like [here](https://github.com/mlx-graphs/mlx-graphs/blob/b0b0aa5be346ce02a92b9f63aa026c1a500315fb/mlx_graphs/datasets/qm7b.py#L41C9-L44C88) when it's needed? So that people can install it if they need that dataset.

Also adding a couple of tests would be great!

mlx_graphs/datasets/dblp.py
```diff
@@ -0,0 +1,141 @@
...
+        node_type_idx = mx.array(node_type_idx).astype(mx.int32)
+
+        """
+        Conference nodes don't have features and hence adding
+        it explicitly to a dictionary will not make sense.
+        Either override the property in the class or set attribute separately
+        for conference
+        """
+        conference_nodes = int((node_type_idx == 3).sum().item())
+        print(conference_nodes)
```
francescofarina, line 99: @vinayhpandya I'm not sure I understand what's going on here - is `conference_nodes` used anywhere?

francescofarina (changes requested):

mlx_graphs/datasets/dblp.py
vinayhpandya: yes it is used for getting the edge index involving conference nodes.

francescofarina: got it 👍

francescofarina (approved):
Looks good!

TristanBilot (approved):
Seems good to me! Great job :)

PR #169, opened 2024-06-27
Add new dataset class for HeterographData

vinayhpandya:
Added new dataset class for Heterograph data

1. Type errors in IMDB and DBLP tests are occurring because the 'Dataset' placeholder is being used to represent both 'HeteroGraphData' and 'GraphData' types.`
2.  Replaced IMDB and DBLP datasets with HeteroGraph dataset to resolve the type errors

## Proposed changes

Please include a description of the problem or feature this PR is addressing.
If there is a corresponding issue, include the issue #.

## Checklist

### DIFF
--- a/docs/source/api/datasets/index.rst
+++ b/docs/source/api/datasets/index.rst
@@ -11,6 +11,7 @@ Datasets
 	:template: dataset-class-template.rst
 
 	Dataset
+	HeteroDataset
 
 
 .. autosummary::
--- a/mlx_graphs/data/data.py
+++ b/mlx_graphs/data/data.py
@@ -1,4 +1,4 @@
-from typing import Literal, Optional, Union
+from typing import Any, Literal, Optional, Union
 
 import mlx.core as mx
 import numpy as np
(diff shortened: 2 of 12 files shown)


francescofarina (changes requested):
I believe this needs a bit more refactoring. We probably want to have a base dataset class from which both `Dataset` and `HeteroGraphDataset` inherit, which contains the shared logic and abstract methods. Right now the `Dataset` class can still accept `HeteroGraphData` objects as inputs, which shouldn't be allowed. @vinayhpandya what do you think?

@vinayhpandya

vinayhpandya:
I think its a good Idea, we can keep `process`,`load` and download function in the parent class and let dataset and heterographdataset class inherit those. @francescofarina what do you think?

francescofarina:
Yes, I believe the parent class could also handle _most_ of the property methods though (like the name, the paths and the number of graphs)

TristanBilot (approved):
Seems OK to me!

PR #174, opened 2025-08-04
bump mlx_version and remove mlx_cluster dependency

### LINKED ISSUE #173: Update to latest mlx version
There's libraries such as `mlx-embeddings` that use a later mlx version that this library would work well with for certain problems that use embeddings alongside transformers. 

I am currently trying to get mlx_cluster to work in the latest MLX C++ api https://github.com/vinayhpandya/mlx_cluster/issues/1 and running into some problems.

francescofarina:
## Proposed changes

This PR bumps the mlx version and removed the mlx_cluster mandatory dependency as it requires mlx==0.18

closes #173

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,12 +17,11 @@ classifiers = [
     "Programming Language :: Python",
 ]
 dependencies = [
-    "mlx==0.18.*; sys_platform == 'darwin'",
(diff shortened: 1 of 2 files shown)


namgyaaal:
I was able to build this fine (with and without uv) and the tests passed on my end.

Issue #172, opened 2025-08-04
Problem building from latest repo due to mlx_cluster


namgyaaal:
**Describe the bug**
Dependency mlx_cluster requires `mlx==0.18.0` which is not listed on the mlx pypi and thus won't allow `pip install -e .` to work.

```
ERROR: Could not find a version that satisfies the requirement mlx==0.18.0 (from versions: 0.0.4, 0.0.5, 0.0.6, 0.0.7, 0.0.9, 0.0.10, 0.0.11, 0.1.0, 0.2.0, 0.3.0, 0.4.0, 0.5.1, 0.6.0, 0.7.0, 0.8.1, 0.9.1, 0.10.0, 0.11.1, 0.12.2, 0.13.0, 0.13.1, 0.14.1, 0.15.2, 0.16.3, 0.17.3, 0.18.1, 0.19.3, 0.20.0, 0.21.1, 0.22.1, 0.23.2, 0.24.2, 0.25.2, 0.26.1, 0.26.2, 0.26.3, 0.26.5, 0.27.1.dev20250725, 0.27.1)
```

**To Reproduce**
Get a .venv with dependencies, git clone and test `pip install -e .` on it. 

**Expected behavior**
Should install fine. 

**Desktop (please complete the following information):**
 - OS version: MacOS Sequoia 15.6
 - mlx-graphs version: Latest from git 
 - mlx version: It requires a version that isn't around anymore

namgyaaal:
Running pip install on a fresh venv brings out the same error. I tested the PR and it compiles fine. Currently running test suite.

francescofarina:
if the bug in `mlx_cluster` gets solved we can add that dependency back in.

vinayhpandya:
Made changes in the API, will release a new version to fix mlx_cluster and verify if it works correctly with mlx_graphs
commit: https://github.com/vinayhpandya/mlx_cluster/commit/bd01e54d6f47dd494f63dc0ef905f64f84a406e6

PR #177, opened 2025-08-14
Use new MLX cluster

vinayhpandya:
**Motivation**
Currently Node2vec is failing because of new version of mlx Utililizing new version of mlx_cluster (along with new api) is recommended

## Proposed changes
Changed mlx_cluster API , documentation: https://vinayhpandya.github.io/mlx_cluster/functions.html#mlx_cluster.random_walk
Changed Node2vec algorithm to evaluate data before sending to random_walk
(Data is not available until evaulation)
To Fix issue: #172

## Checklist


mlx_graphs/algorithms/node2vec.py
```diff
@@ -53,6 +53,7 @@ def __init__(
         p: float = 1.0,
         q: float = 1.0,
         num_negative_samples: int = 1,
+        use_gpu: bool = False,
```
francescofarina, line 56: why give the option here? shall it just be run on gpu always?

francescofarina:
@vinayhpandya building mlx_cluster currently fails for me

mlx_graphs/algorithms/node2vec.py
vinayhpandya: I wanted to benchmark both CPU and GPU times in node2vec notebook, I can change it to use GPU by default

francescofarina:
> Are you trying to build mlx_cluster from source or are you just installing from pyPI

Installing from PyPi. I tried multiple versions of python, both with pip and uv and always get the error above. Can you try installing it in a clean environment?

mlx_graphs/algorithms/node2vec.py
vinayhpandya: Updated

francescofarina (approved):

